PR #135, opened 2014-10-26
Add info about intermadiate CA

mat-mo:
rapidssl is a good example of a CA that requires a chained crt


M5oul:
https://yunohost.org/#/certificate_en has been update since.

PR #153, opened 2014-11-28
h5ai a modern HTTP web server index

Gastlag:
h5ai is a modern file indexer for HTTP web servers with focus on your files. Directories are displayed in a appealing way and browsing them is enhanced by different views, a breadcrumb and a tree overview. Initially h5ai was an acronym for HTML5 Apache Index but now it supports other web servers too.

See the demo directory with most features enabled : http://larsjung.de/h5ai/demo/

Works fine with Apache httpd, lighttpd, nginx and Cherokee.


M5oul:
Done with https://github.com/YunoHost/doc/commit/3c23b03c471530c000612127b22196ab506f57d5

PR #236, opened 2015-05-13
Update admindoc_fr.md

Jean31:
Add an entry in TOC for DNS resolver


M5oul:
https://github.com/YunoHost/doc/commit/6edfc512068f131a294f52d7616654ad754132fa. I prefer maintained only one page. It give me less works…

PR #237, opened 2015-05-13
Update admindoc.md

Jean31:
Translate to english


M5oul:
Bad choise to translate the french page. It better to take the [english sitemap](https://yunohost.org/#/sitemap_en). I will think to delete this admin, user and contributor pages and maintain only Sitemap.

PR #283, opened 2015-06-08
Two tiny changes

evotopid:
I just corrected two little things that I came across while reading the docs.


Kloadut:
This page is pooly translated anyway.

But thanks! :-)

PR #287, opened 2015-06-28
Use of mail-tester.com instead of check-auth@verifier.port25.com

polytan02:
check-auth@verifier.port25.com seems to be old and deprecated


M5oul:
Done with https://github.com/YunoHost/doc/commit/fff0eb5e9777355d484ed2573e4cc80b33df2b1f.

PR #289, opened 2015-06-28
check-auth@verifier.port25.com is deprecated

polytan02:
Use of mail-tester.com instead of check-auth@verifier.port25.com which is deprecated.


M5oul:
Done with https://github.com/YunoHost/doc/commit/c5abdf2cdfda233be6ccd24f227024fed0eb0693.

PR #291, opened 2015-06-28
Add ip4 and ip6 in SPF

polytan02:
Add of ip4 and ip6 fields in SPF
Comment indicating that ip6 section is not compulsory and can be removed


M5oul:
Done with https://github.com/YunoHost/doc/commit/65c67c4cb6eebe37ebe1f25b2f3f006f71a067d3.

PR #299, opened 2015-08-03
Update apps_wishlist.md

laskameleonart:
Proposition to add Kontalk to the Apps wishlist. Kontalk is a fully open source (both client and server) short message service featuring encrypted communication. The server side could be integrated to YunoHost.

Hope others will like the idea :-)


M5oul:
This seems to be an application for Android mobile. Is there a web interface like Jappix? For server side, YunoHost already come with a XMPP instant messaging server called Metronome.

laskameleonart:
What makes Kontalk interesting is it's something like WhatsApp, an instant messaging / sms like App for smartphones. But instead of being proprietary like WhatsApp, both the client and server side are FOSS. So if you install your own server, you have your own free private SMS App for you and your friends (or everyone you accept on your server).

It's very different from XMPP (and i think they are complementary).

PR #313, opened 2015-09-08
add wildcard in dns config example

Psycojoker:
Ohai,

Currently YUNOHOST doesn't recommand to add a wildcard to the domain name which is a bit sad since it doesn't cost anything and allow more option to the user. Here is a patch to fix this.

Additionnal coment: apparently 900 for a TTL value is way to low and should be more something like 21600 according to PEB on #ffdn

Cheers :heart:

### DIFF
--- a/dns_config.md
+++ b/dns_config.md
@@ -5,7 +5,9 @@ Sample DNS config for domain.tld
 ```bash
 #### IPv4 & IPv6
 @ 900 IN A 111.222.333.444 (Minimal)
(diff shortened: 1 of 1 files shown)


opi:
+1 for setting a wildcard, or at least explain its benefits.

About the TTL, I don't know where this value came from, and we could certainly increase it (in another PR ;)

Kloadut:
zupa !

PR #321, opened 2015-09-22
Adding Debian 8 in prerequisite in addition to Debian 7

paoesco:
Yunohost works on Debian 7 and Debian 8. Adding Debian 8 on pre requisite

### DIFF
--- a/install_on_debian.md
+++ b/install_on_debian.md
@@ -8,7 +8,7 @@
 
 An ARM box, a VPS, a dedicated server, a standard x86 computer, an old Macintosh, ...
 
-* with **Debian 7** (Wheezy, Raspbian, Cubian, etc.) installed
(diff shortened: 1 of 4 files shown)


M5oul:
Done. Thanks.

opi:
Why not ask for Debian Jessie (8) only for now ? I mean, it's current stable ;)

M5oul:
We could recommend Jessy and say YunoHost also support Wheezy.

opi:
you're right, let's do that !

( btw, repeat after me : jessie, jessie, jessie ... )

M5oul:
In fact, I thought it was jessy in english!

opi:
https://en.wikipedia.org/wiki/Jessie_%28Toy_Story%29

PR #323, opened 2015-09-25
Add dolibarr erp crm into wishlist doc

eldy:


M5oul:
Ever done :smiley_cat:

PR #331, opened 2015-10-24
Correcting misspellings on installation pages

drkokandy:
just a couple misspellings I noticed when installing.


Kloadut:
Thanks !

PR #337, opened 2015-11-20
Add Emby, Deluge and Sonarr to wishlist

franga2000:
Must-have apps for a home media server. 
And I know deluge is old, but it's still the best client out there.


M5oul:
Done with https://github.com/YunoHost/doc/commit/3138bd272ee7aa64220b79f285ce915755d29e3f.

PR #346, opened 2016-01-13
Add Shout IRC to apps Wishlist

twodimes:


mbugeia:
Shout is already packaged https://forum.yunohost.org/t/shout-a-cool-irc-web-client/823

PR #376, opened 2016-05-31
proposal of a new image docker yunohost

aymhce:
require a host machine with systemd and the docker image must be launched with the privileged mode

### DIFF
--- a/docker.md
+++ b/docker.md
@@ -57,7 +57,8 @@ Please follow *Building* section [here](https://github.com/aymhce/YunohostDocker
 
 To start the container, run the next command by replacing DOMAIN by a valid domain e.g.: example.com => yunohost.example.com
 ```bash
-docker run -h yunohost.DOMAIN -v $(pwd):/yunohost -d --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro domainelibre/yunohost:2.4 /bin/systemd
+docker run -h yunohost.DOMAIN -v $(pwd):/yunohost -d --privileged \
+-v /sys/fs/cgroup:/sys/fs/cgroup:ro domainelibre/yunohost:2.4 /bin/systemd
 ```
 
 If you want to run the container and forward all the interesting ports to the host:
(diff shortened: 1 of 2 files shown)


Psycojoker:
This PR is a bit dying alone, it's sad :(

docker.md
```diff
@@ -26,32 +29,42 @@ sudo pacman -Sy docker
...
+
+Please follow *Building* section [here](https://github.com/aymhce/YunohostDockerImage)
 
 ---
 
 ## Run the container
+
+> **The linux docker host must run systemd**
+
+> **WARN : --privileged mod is required for systemd**
```
zamentur: YunoHost should detect it is install in docker to disable firewall operation. Without that if you run --priviledge, each app or domain install will redefine firewall rules of the host.

For me it is a very important point  to fix. Without I am against to speak about an official way to run yunohost with docker.

alexAubin:
Unfortunately I kind of agree with @zamentur : if Docker's --privileged is the same as LXC's, then they are [considered unsafe](https://linuxcontainers.org/lxc/security/). While this may be suitable for tests (assuming you trust the software you're testing), it's not for production : you could imagine an app shipping some nasty code to take over the *host* of the container.

So, not sure what to do about this... While it's cool to have some documentation and an image, we can't really advertise about it. Or if we do publish this, we should clearly state it's considered unsafe for prod.

scith:
Couldn't we create 2 docker containers instead: one volume only and one for YNH.
This way, we won't have to run the YNH container in privileged mode (I think). We can just destroy the YNH container and recreate it with extra ports opened, without losing any data.
The volume container could eventually link the guest home folder (or yunohost.backup) to the host.

?

alexAubin:
Personally I don't understand the use case and don't realize wether there's huge demand for a docker image... It looks like this require some work to have something safe that can officially support / document ...

In the meantime, since there's not much activity on this PR, I'm flagging this as inactive. Feel free to speak up if you care / know about / want to maintain this.

PR #382, opened 2016-07-04
Add warning about 'reset admin password' method.

opi:


jeromelebleu:
Oh, I didn't noticed, what didn't worked?

opi:
First, there is no more `/etc/slapd/slapd.conf` on jessie, and then, playing with `/etc/ldap/slapd.conf` does not allow me to log with a `slappasswd` defined password.
I add this warning to doc pages while we find a clean way to enforce a YunoHost admin password.

opi:
I'm not alone experiencing this issue: https://forum.yunohost.org/t/reset-admin-password/1681

jeromelebleu:
Indeed, there are some mistakes and the procedure is incomplete...

> First, there is no more `/etc/slapd/slapd.conf` on jessie

You're right, the right file is `/etc/lapd/slapd.conf`- see the configuration regeneration for slapd [here](https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/conf_regen/06-slapd#L62).

> playing with `/etc/ldap/slapd.conf` does not allow me to log with a `slappasswd` defined password.

This file is only used to generate the `/etc/ldap/slapd.d` directory - which is used by LDAP. We should add in this how-to that it must be regenerated after that. One way to do that is to execute the following:

```bash
bash /usr/share/yunohost/hooks/conf_regen/06-slapd init
```

opi:
Even when playing with `06-slapd init`, I can't reset the LDAP admin password.

jeromelebleu:
Oops, sorry, I forgot that this script replaces the `slapd.conf` file with the regular one... Any modifications are lost after executing it.

Here are instead the commands to do it by hand:

```bash
# optional: check the  config file at first
slaptest -Q -u -f /etc/ldap/slapd.conf

# regenerate LDAP config directory from slapd.conf
rm -Rf /etc/ldap/slapd.d
mkdir /etc/ldap/slapd.d
slaptest -f /etc/ldap/slapd.conf -F /etc/ldap/slapd.d/ 2>&1
chown -R openldap:openldap /etc/ldap/slapd.d/
```

PR #383, opened 2016-07-16
Port forwarding: add port 587 for SMTP with SSL

likeitneverwentaway:
Spent way too many hours figuring out why I couldn't send any mail...


M5oul:
Thanks.

PR #388, opened 2016-08-28
Pentest documentation

M5oul:
I added pentest links in security part of documentation.

That could show how secure is YunoHost and what have been done on it.

Are you ok to publicly display it?

### DIFF
--- a/security.md
+++ b/security.md
@@ -118,6 +118,6 @@ sudo service yunohost-api stop
 
 Some [pentests](https://en.wikipedia.org/wiki/Penetration_test) have been done on a YunoHost 2.4 instance (french):
 
-- [#1 – Preparation](https://blog.exadot.fr/2016/07/03/pentest-dune-instance-yunohost-1-preparation)
-- [#2 – The functionning](https://blog.exadot.fr/2016/07/12/pentest-dune-instance-yunohost-2-le-fonctionnement)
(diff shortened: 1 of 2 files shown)


Psycojoker:
Sound good (but I'm not 100% sure about it).

Regarding security since this is the subject: we should publish our security@yunohost.org somewhere with it's pgp no? (And sign that part of the page using this key to prove its us).

@opi: since you have created this key, what's your opinion on that? ^

M5oul:
We could wait for last part of pentest (Grey Box) before signing something.

Psycojoker:
Hm why? This gpg key is a way for people to communicate to us in a secure way in which they know that we will be the only ones able to read what they've said, this is a secure way to send us a email saying "hey, here is a critical security vulnerability and how to fix it".

M5oul:
I am talking about:
>(And sign that part of the page using this key to prove its us)

Psycojoker:
> I am talking about:
>> (And sign that part of the page using this key to prove its us)

Hm, why should we wait for the end of the pentesting then? I'm not sure to guess the reason (as this is decorated from it, it's just a way to prove that we are the owners of this PGP key).

M5oul:
Ah, I thought you were talking about signing `security*.md` pages with GPG key.

Psycojoker:
Ah no, just the extract with the key ;)

PR #402, opened 2016-10-07
Selfhosting en

theodorefaure:
progress on “install” translation

### DIFF
--- a/install_on_arm_board.md
+++ b/install_on_arm_board.md
@@ -7,9 +7,9 @@
 <img src="/images/cubieboard2.png">
 <img src="/images/micro-sd-card.jpg">
 
-* An ARM board with 500MHz CPU and 512Mo RAM.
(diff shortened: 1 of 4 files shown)


install_on_arm_board.md
```diff
@@ -38,4 +38,4 @@
 
 ---
 
-***If you need help during one of these steps, do not hesitate to use [our support tools](/support).***
+***If you need help during one of these steps, do not hesitate to use [our support channel](/support).***
```
M5oul: I prefer `tools` as there this also the forum.
Same on the other page.

theodorefaure: It was meant as channel in the sense of "canal" and not as "chat". I agree this can be confusing. What do you think of "You can get support if you need help during one of these steps" or "Need help during one of these steps? Get support!"

M5oul: Yep, it's better!

PR #430, opened 2016-12-27
[enh] add architecture schema

Psycojoker:
Hello,

I've finally ported my whiteboard drawing to a cool schema. This as been done using https://draw.io (freesoftware normally) and the xml file is the export of draw.io if we want to modify it.

Result:

2016-12-27-222644_1149x729_scrot


Psycojoker:
Ping @opi :)

Psycojoker:
Ah btw, I don't know where to put that in the doc so I'm going to let you chose that.

@opi you've also made a schema but I couldn't find your slides :/ do you know where it is?

PR #437, opened 2017-01-17
add Kresus on wishlist

Snutin:


scith:
Would be nice :) I haven't managed to package it yet and don't gave time but feel free to take it forward scith/kresus_ynh

PR #441, opened 2017-01-18
[fix] Update doc about vagrant setup

zamentur:


Psycojoker:
Shouldn't we recommend at the beginning of the file to use `ynh-dev` instead?

Psycojoker:
Thanks :)

PR #442, opened 2017-01-18
[enh] Add a documentation dedicated to olinuxino

zamentur:
Translation in english need to be done


install_on_olinuxino.md
```diff
@@ -0,0 +1,106 @@
...
+
+Le matériel nécessaire :
+* Un mini-serveur Olimex :
+ * [A20-OLinuXino-LIME](https://www.olimex.com/Products/OLinuXino/A20/A20-OLinuXino-LIME/open-source-hardware)
+ * [A20-OLinuXino-LIME2](https://www.olimex.com/Products/OLinuXino/A20/A20-OLinuXino-LIME2/open-source-hardware)
+* Une carte micro-SD (des [Transcend 300x](http://www.amazon.fr/Transcend-microSDHC-adaptateur-TS32GUSDU1E-Emballage/dp/B00CES44EO) pour des raisons de performance/stabilité).
+* Un adaptateur secteur [européen](https://www.olimex.com/Products/Power/SY0605E/) pour alimenter la carte olimex. L’alimentation via USB semble peu stable.
+* Un câble Ethernet/RJ-45 pour brancher la Brique à son routeur.
+
+Et évidemment, **un ordinateur sous GNU/Linux ou BSD**.
```
alexAubin: This note was already there in previous version, but it's not quite friendly for people running Windows :/ ... Flashing an image on an SD card on windows is already documented [here](https://yunohost.org/#/copy_image) so we could point Windows user to it. Don't know about arp-scan but I think the command exist on Windows as well (even though Windows's CLI is crap). Then we'd need to tell people to use PuTTy (or some other client if anybody knows one).

But we can add this later if nobody feels like documenting it.

M5oul:
Listing on sitemaps is missing.

PR #446, opened 2017-01-24
Add Let's encrypt image and text

M5oul:


Psycojoker:
I don't see how this PR could generate disaproval so I'm passing it as a micro decision. Feel free to revert if needed.

PR #464, opened 2017-02-19
[enh] packaging_apps_manifest: add 'version' field.

M5oul:
I will add this test on the linter.
This field is mandatory, may I add it on the linter?
It would be great to have this check as medium issue.

### DIFF
--- a/packaging_apps_manifest.md
+++ b/packaging_apps_manifest.md
@@ -12,7 +12,7 @@ The `manifest.json` file defines the app's constants, a bunch of values that Yun
         "fr": "Webmail Open Source"
     },
     "url": "http://roundcube.net/",
-    "version": "v1.0",
+    "version": "1.0-1",
     "license": "free",
     "maintainer": {
         "name": "kload",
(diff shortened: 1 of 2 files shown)


Psycojoker:
Do we take this field into account? I don't remember seeing it anywhere.

M5oul:
YunoHost does nothing about this version.
It could be used for many reasons.
It could be interesting to have it.

Psycojoker:
Ah, so if you want to modify the manifest and had a new field you need to call for a decision, manifest modifications really needs to be discussed since this will affect all app developers.

I'm not against this modification (quite the opposite actually) but this need to be done properly and even integrated into YunoHost/apps page would be good and we'll need to communicate to the app developers for this modification.

packaging_apps_manifest.md
```diff
@@ -62,6 +63,8 @@ The `manifest.json` file defines the app's constants, a bunch of values that Yun
 
 * **url**: software website.
 
+* **version**: version of upstream software which is installed. Must be a string.
```
zamentur: Personnaly i don't put the axact upstream version number. Because I used it to manage the upgrade procedure of the package. So, in my case it is the package version number, I create it like this:
UPSTREAM_VERSION_NUMBER-PACKAGE_NUMBER

alexAubin:
Bump - shall we implement @zamentur's comment ?

alexAubin:
Bump :/ ... What do we do with this ?

JimboJoe:
Should we have this systematic suffix or could we just skip it for the first upstream version packaging i.e. 1.0, THEN 1.0-2 (or 1...), etc. ?

alexAubin:
@JimboJoe I have no strong opinion on this but imho the suffix should be systematic

alexAubin:
Ok guys, merging this

PR #467, opened 2017-02-25
Debian VPS installation feedback

Djiko:
Hi all,

a couple of update proposals after a fresh installation on a Debian VPS. I run into 2 issues and did not immediately find the solution. This might help others.

Keep up the good work.

### DIFF
--- a/install_manually.md
+++ b/install_manually.md
@@ -8,7 +8,7 @@ Once you have access to your server, either directly or by SSH, you can install 
 
 1. Install git
 ```bash
-sudo apt-get install git
+sudo apt-get install git dialog
 ```
 
 2. Clone the Yunohost install script repository
--- a/install_manually_fr.md
+++ b/install_manually_fr.md
@@ -8,7 +8,7 @@ Une fois que vous avez accès à votre serveur, directement ou par SSH, vous pou
 
 1. Installez git
 ```bash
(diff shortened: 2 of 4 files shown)


install_manually.md
```diff
@@ -8,7 +8,7 @@ Once you have access to your server, either directly or by SSH, you can install
 
 1. Install git
 ```bash
-sudo apt-get install git
+sudo apt-get install git dialog
```
M5oul: Why do you propose `dialog` package?
Can't we install it from install script or YunoHost debian dependency?
Installing `git` is only to clone `install_script` repository.

M5oul:
The rest of the modification is ok for me.

Djiko:
Dialog is used to display windows such as this one in the shell

https://yunohost.org/images/install_script.png

If it's not installed, you don't see anything, the installation process 
doesn't hang but doesn't process either as it is not able to display the 
needed windows. Debian VPS might not have it installed.

This is the way I solved it (easier for me, did not have to touch the 
install script at all), but as the install script already 
installs/updates some packages, this is certainly the best way to go.

install_manually.md
M5oul: Ok, I didn't found where to install `dialog` on the install script.
I am ok to install it here.

PR #469, opened 2017-03-06
Recommend to use the image when installing on RPi + update manual install

alexAubin:
On the support and on the forum, I see many user desperately trying to manually install from Raspbian and having trouble with it. The RPi image aims to make things much simpler and therefore we should recommend it.

### DIFF
--- a/install_on_raspberry.md
+++ b/install_on_raspberry.md
@@ -12,14 +12,16 @@ Before setting up a server at home, it is recommended that you know the [possibl
 
 ## Pre-requisite
 
+- A Raspberry Pi 0, 1, 2 or 3 ;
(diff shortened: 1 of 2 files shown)


install_on_raspberry.md
```diff
@@ -31,33 +33,28 @@ Before setting up a server at home, it is recommended that you know the [possibl
 
 ---
 
-## Manual installation
+## Manual installation (advanced users)
+
+<div class="alert alert-info" markdown="1">
```
Psycojoker: I would personally used `alert-danger` but I'm not known for my subtle style.

alexAubin:
I propose to merge this in a few days if there's no more comments

PR #472, opened 2017-03-17
add apps

Thatoo:
Add ONLYOFFICE Document Server, Collabora Online server, Collabora Online, LibreOffice Online


M5oul:
This is big applications which uses Docker and Node.js.
It could be packaged but won't be easy.

PR #473, opened 2017-03-19
[enh] migration from Metronome to Prosody

M5oul:
- change certificate group to 'prosody'.
- update logo.

**To be merged at the same time as https://github.com/YunoHost/yunohost/pull/240**


frju365:
As for instance we won't move to Prosody, can we close this PR and reopen when needed ?

kay0u:
Bump

frju365:
I close this PR for instance, as metronome is still running. If opposition, I think first we must finish (if needed) PR in core, and then ok (the project has decided to keep working with metronome for instance as long as it's maintained).

PR #480, opened 2017-04-07
[mod] remove '-p' from release version notation.

M5oul:
It's confusing on documentation.
I put `2.6.0-p` and it's not what I wanted.
I have to restart with `2.6.0`.
`-p` is used for what?
Can we remove it?
If it's usefull, we may display an other place with explanations.


M5oul:
@opi @Psycojoker ?

PR #514, opened 2017-08-05
Add solution when email can't be send

artonge:
As seen in https://forum.yunohost.org/t/impossible-denvoyer-des-mails-avec-roundcube/1698/16?u=artonge

### DIFF
--- a/email_fr.md
+++ b/email_fr.md
@@ -31,6 +31,11 @@ De nombreux messages permettant d’identifier les problèmes se trouvent enregi
 
 <!-- ajouter une doc pour consulter ses logs des services mail -->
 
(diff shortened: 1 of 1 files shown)


frju365:
Can you add these different links at the end of the solution ? (In fact it's information per ISP/FAI) Because in fact, not all the box have the port 25 blocked.
add : 
```
### More Info ###
https://yunohost.org/#/isp_fr
https://yunohost.org/#/isp_orange_fr
https://yunohost.org/#/isp_sfr_fr
https://yunohost.org/#/isp_free_fr
```

artonge:
I don't have access to the old repo anymore. See #633

PR #537, opened 2017-08-28
Display the list of apps with nice cards

zamentur:
At the beginning the goal was to emphasize the "Documentation" button.

I have a bit gone of topic ... :p 

So here are the result :)

app_list
app_list_2
app_list_3


frju365:
I'm not really fan of scripts (JS-CSS) inside md document... Is it possible to do two external files style.css et script.js.

Moreover, I had the idea of adding the link of the latest official test of each apps.

zamentur:
> I'm not really fan of scripts (JS-CSS) inside md document... Is it possible to do two external files style.css et script.js.

The goal was to allow to change easily in Simone. And I have not a personnal instance of Simone.

But I am not satisfied too.

There is also some bad css structures (for example the width of each card is fixed due to issue I have encouter with the bootstrap grid, and the description could be too long and go over the update time...).

But it is already better than before.

PR #591, opened 2017-10-24
Add page for framasoft apps equivalence

alexAubin:
Proposal to add this page which should be easier to read and maintain compared to the Redmine roadmap and the forum thread... Though maybe we should have a better display (like the /apps page).

2017-10-24-210250_1366x768_scrot


frju365:
Peut-on merger si tout le monde est d'accord ?

ldidry:
ok pour moi

ldidry:
Nonobstant le framalist**e**s

PR #633, opened 2018-01-14
 Add solution when email can't be send (bis)

artonge:
As seen in https://forum.yunohost.org/t/impossible-denvoyer-des-mails-avec-roundcube/1698/16?u=artonge

:warning: Duplicate of #514. I didn't add access to the old repo so I create a new one and created this PR with the according changes you asked.


frju365:
Thanks, @artonge & @alexAubin !

PR #671, opened 2018-02-27
Create install_ar.md

BoFFire:


BoFFire:
Oups /o\ some typo will we corrected in v2.

PR #705, opened 2018-04-17
[enh]  Use KiwiIRC instead of XMPP

Rafi594:
Not ready yet.


Rafi594:
I'll work on the kiwiirc app for install it on the infra instead of XMPP.

frju365:
Ok. Let's merge it now... ! Anyway, we need to merge it now.

PR #718, opened 2018-05-08
Add a page to list pre-installed images

alexAubin:
So this is an attempt to provide a replacement for the html at build.yunohost.org that list available pre-installed images.

The motivation behind this is that : 
- having yet another individual page not in the "standard" doc is inconsistent in term of UX, and has a maintenance cost
- we now have automatic / semi-automatic image builds and often need to update the links on this page. Currently this is done in a dirty way
- there are plans to produce images for many ARM boards in the mid-term, and it's not trivial to integrate those in the current page

In this proposal, the list of available images is kept in a json here : http://build.yunohost.org/images.json (which can be updated by scripts easily), and "cards" are created in a similar way than the apps page (https://yunohost.org/#/apps)

We get the following : 

2018-05-08-235553_1366x768_scrot

### DIFF
--- a/images.md
+++ b/images.md
@@ -1,0 +1,114 @@
+# Pre-installed images
+
+<div id="cards-list">
+</div>
+
+<script type="text/template" id="image-template">
+<div id="{id}" class="card panel panel-default">
+        <div class="panel-body text-center">
+            <h3>{name}</h3>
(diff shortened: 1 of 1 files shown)


frju365:
a third review pls ?

frju365:
ok. Let's merge.

PR #736, opened 2018-06-04
Update contributor doc and clean a few outdated pages

alexAubin:
Similar work to https://github.com/YunoHost/doc/pull/735 and https://github.com/YunoHost/doc/pull/734, this time trying to clean contributordoc.md and related pages.

After updating it I realize I should update the french version too and I was like, meh, do we really need to maintain a french doc meant for contributors, considering that we already have a hard time maintaining the english version ... So for now I propose to remove the french version :| 

About the build_* pages, those are now handled on corresponding git repo's README or internally...


frju365:
Remove the french version is a bad thing from my point of view. I can maintain it, if you want, but I'm against removing it.

frju365:
For the other part, it's totally ok.

PR #759, opened 2018-06-23
[enh] add ynh.fr in documentation about the dns

frju365:


frju365:
I merge as we discuss of it in the previous PR #758

frju365:
Next step is to do the translation in DE and PT.

PR #812, opened 2018-09-17
Added Njalla to List of Doman Registrars

0verk1ll:


alexAubin:
Hellow,

we discussed this in yesterday's meeting and concluded that a note should be added about the fact that the person won't have full control/ownership of the domain. Does that sound alright to you ?

PR #822, opened 2018-10-09
Add Retroshare and a Scuttlebutt pub to wishlist

amuuza:
Add Retroshare and a Scuttlebutt pub to the app wishlist.


frju365:
Thanks. !

PR #868, opened 2018-11-27
Started italian translation

silkevicious:
Translated some files into italian.
Hope this could fix this: https://github.com/YunoHost/Simone/pull/12#issuecomment-441739264
because even with disable italian in the Simone menu, i still get the main page void.


alexAubin:
Thanks a lot ! I don't speak italian but I trust you :P 

Will reenable italian in Simone, please complain if that still doesn't work

PR #898, opened 2019-01-13
backup_core_only option to backup nextcloud data

GohuHQ:
I have been trying to backup nextcloud with "backup_core_only -v 0" but it didn't work until i found a wiki somewhere on the internet which mentions "backup_core_only -v ''" and my data backup worked only since.
I hope it helps.

### DIFF
--- a/backup.md
+++ b/backup.md
@@ -54,7 +54,7 @@ For more informations and options about backup creation, consult `yunohost backu
 
 #### Apps-specific configuration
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Uh, do confirm, @maniackcrudelis or anyone from @YunoHost/apps  who knows about this ?

maniackcrudelis:
https://github.com/YunoHost-Apps/nextcloud_ynh/blob/master/scripts/backup#L82
Anyway, a regular backup, aside from a backup during a upgrade, shouldn't use this option. So all data should have been backup.

Also, nextcloud shouldn't use anymore that syntax, but rather use is_big argument from backup helper and BACKUP_CORE_ONLY global variable.
Therefore, nextcloud have to be updated...

alexAubin:
Merging because people are indeed encountering this issue

PR #899, opened 2019-01-13
Backup workaround

Gofannon:
Add a workaround for people to get an access to their backups with the **admin** user

Should be fixed in the core. See commit details

PR #900, opened 2019-01-13
[enh] Add graphical program to exchange files

Gofannon:
Add graphical program to exchange files with Yunohost on major PC OS
Depends on https://github.com/YunoHost/doc/pull/899 for the page `backup_workaround.md`

----

It think that it should not be limited to backups but to manage other apps like music in https://github.com/YunoHost-Apps/sonerezh_ynh. Or nextcloud can do the trick?


alexAubin:
Merging, assuming we're good with this. Thanks again !

frju365:
Well for me, nothing to say : It's very good. Just needed translation ;-) Thanks you both ! :-)

PR #927, opened 2019-02-20
Update packaging_apps_levels.md

maniackcrudelis:
According to our discussion during the last meeting.
We would like to use the level 8 for high quality packages.

So I made a little refactoring of the levels. Also I change the level 4, which is quite complicated to handle and not really representative of the real level of an app.

To explain my modifications:
- The level 4 is now about how the app handles backup and restore. It's a must have now.
- While the level 6 is, for the automatic part with Package_check, is only about the app being in YunoHost-Apps.
Not that much, but didn't find anything else we could check automatically without changing all levels.
And anyway, nowadays, we're much about very low level (1-3) and level 7.
- Level 7 gets many YEP that were in level 8 and 9, but actually already expected for a level 7 app.
- Level 8 gets ldap and SSO. Since we can't check that automatically.
- Level 9 gets app isolation. We don't know yet how to do that properly for most of our apps

The level 8 would be now a manual level (Accorded only by Apps group) about high quality apps that respect all packaging recommendations.

Related to https://github.com/YunoHost/apps/pull/677

**This PR should be merged after https://github.com/YunoHost/doc/pull/935**

### DIFF
--- a/packaging_apps_levels.md
+++ b/packaging_apps_levels.md
@@ -23,22 +23,22 @@ The application installs and uninstalls correctly in all common configurations.
 The application can be updated.
 
 **Level 4**
-The application uses YunoHost users directly and allows unique identification from the YunoHost portal.
+The application can be saved and restored.
 
 **Level 5**
 The application package code follows some syntax rules.
 
 **Level 6**
-The application can be saved and restored.
+The application package is in the YunoHost-Apps organization.
 
 **Level 7**
 The application package passes all integrity tests successfully.
 
 **Level 8**
-The application respects a set of advanced recommendations improving its general quality.
+The application package respects all packaging recommendations. This is a high quality app.
 
 **Level 9**
-The application complies with all recommendations. This is an excellent quality package.
+The application complies with higher packaging recommendations. Not available yet.
 
 **Level 10**
 The application package is considered perfect!
@@ -88,7 +88,6 @@ YEP to be respected to reach level 2:
 - *[YEP 2.18.4](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-2184---supporter-linstallation-sur-un-sous-dossier----valid%C3%A9--auto--official-) : Support installation on a subfolder*
 - *[YEP 4.6](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-46---g%C3%A8re-le-multi-instance----valid%C3%A9--manuel--optional-) : Manage multi-instance*
 
-
 ### Level 3
 
 **The application supports upgrade from an older version of the package.**  
@@ -100,16 +99,11 @@ YEP to be respected to reach level 3:
 
 ### Level 4
 
-**The application supports LDAP and HTTP AUTH**
-
-The application manages its users directly from the [YunoHost ldap base](https://github.com/YunoHost/SSOwat/blob/366dd6c4438e6550f7438c36893690b628340185/config.lua#L50-L53) and allows unified connection using [HTTP authentication](https://fr.wikipedia.org/wiki/Authentification_HTTP) from the SSO.
-
-*If the application is not capable of supporting an ldap directory or HTTP authentication, this level can be ignored.  *
-*However, it is necessary to be able to justify this impossibility*.
+**The application can be backed up and restored without error on the same machine or another.** 
 
 YEP to be respected to reach level 4:
-- *[YEP 4.1](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-41---lier-au-ldap----valid%C3%A9--manuel--official-) : Link to ldap*
-- *[YEP 4.2](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-42---lier-lauthentification-au-sso----valid%C3%A9--manuel--official-) : Bind authentication to sso*
+- *[YEP 4.3](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-43---fournir-un-script-de-sauvegarde-yunohost-fonctionnel----valid%C3%A9--auto--official-) : Provide a functional YunoHost backup script*
+- *[YEP 4.4](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-44---fournir-un-script-de-restauration-yunohost-fonctionnel----valid%C3%A9--auto--official-) :  Provide a functional YunoHost restore script*
 
 ### Level 5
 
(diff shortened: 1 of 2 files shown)


maniackcrudelis:
@YunoHost/apps

packaging_apps_levels.md
```diff
@@ -23,22 +23,22 @@ The application installs and uninstalls correctly in all common configurations.
 The application can be updated.
 
 **Level 4**
-The application uses YunoHost users directly and allows unique identification from the YunoHost portal.
+The application can be saved and restored.
```
alexAubin: Ogod I am so happy if we can change this

I am just concerned by what's the plan to handle the "migration" for old check_process files with `Level 4=0|na|1`. Do you know how / if the CI can handle that cleanly ? Or shall we edit all check_process files or something ?

maniackcrudelis: I'm going to make a pull request for package_check, it handles that part.
Still, if we indeed can ignore na here, we can't do anything about 1.

We will have to handle it manually

packaging_apps_levels.md
```diff
@@ -151,20 +140,23 @@ If an app reaches level 6, it may qualify to join the official applications.
...
+- *[YEP 2.16](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-216---v%C3%A9rifier-la-disponibilit%C3%A9-des-d%C3%A9pendances-sur-arm-x86-et-x64----valid%C3%A9--manuel--official-) : Check dependency availability on ARM, x86 and x64*
 - [YEP 2.18.5](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-2185---ajouter-la-tuile-yunohost-pour-naviguer-facilement-entre-les-applications----valid%C3%A9--manuel--official-) : Add the YunoHost tile to easily navigate between applications
-- [YEP 3.4](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-34---isoler-lapp----brouillon--manuel--official-) : Isolate app
+- [YEP 4.1](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-41---lier-au-ldap----valid%C3%A9--manuel--official-) : Link to ldap
+- [YEP 4.2](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-42---lier-lauthentification-au-sso----valid%C3%A9--manuel--official-) : Bind authentication to sso
```
Josue-T: I don't think that all actual apps at the (actual) level 7 does both (SSO + LDAP integration). So So maybe it could be good do define more that.

By example I think about discourse or friendica which support only LDAP but not the SSO, So what does we do about that ?

packaging_apps_levels.md
```diff
@@ -175,32 +167,13 @@ If an application is not available on an architecture, and it is impossible to c
...
-- [YEP 2.11](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-211---utiliser-une-variable-plut%C3%B4t-que-lapp-id-directement---valid%C3%A9--manuel--official-) : Use a variable rather than the app id directly
-- [YEP 2.13](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-213---traduire-le-package-en-anglais----brouillon--manuel--official-) : Translate the package into English
 - [YEP 2.14](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-214---remplir-correctement-un-fichier-de-conf----brouillon--manuel--official-) :  Fill a conf file correctly
 - [YEP 2.17](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-217---prendre-en-compte-la-version-dorigine-lors-des-mises-%C3%A0-jour----valid%C3%A9--manuel--official-) : Take into account the original version during updates
+- [YEP 3.4](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-34---isoler-lapp----brouillon--manuel--official-) : Isolate app
```
Josue-T: Maybe fore that we can say that need to run as a specific system user. I know that it's not really "Isolation" but it's more easily possible.

Or by example for python apps we can say that we want that the app is installed in a virtual environment.

It just still some proposition to be more clear about that...

packaging_apps_levels.md
maniackcrudelis: What do you mean ?
If an upstream app does support SSO and LDAP but not the package, then that's a problem that should be fixed.

packaging_apps_levels.md
maniackcrudelis: That's not really isolation indeed.
php provide a way to really isolate apps, but that too effective.

Anyway, I put that on level 9, because we don't have any real solution for the moment.

packaging_apps_levels.md
zamentur: We could add a version format for the checkprocess, like that we could know if level 4 is for sso/ldap or not.

packaging_apps_levels.md
```diff
@@ -175,32 +167,13 @@ If an application is not available on an architecture, and it is impossible to c
 
 YEP to be respected to reach level 9:
 
-- [YEP 2.10](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-210---configurer-les-logs-de-lapplication----brouillon--manuel--working-) : Configure application logs
-- [YEP 2.11](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-211---utiliser-une-variable-plut%C3%B4t-que-lapp-id-directement---valid%C3%A9--manuel--official-) : Use a variable rather than the app id directly
-- [YEP 2.13](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-213---traduire-le-package-en-anglais----brouillon--manuel--official-) : Translate the package into English
 - [YEP 2.14](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-214---remplir-correctement-un-fichier-de-conf----brouillon--manuel--official-) :  Fill a conf file correctly
 - [YEP 2.17](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-217---prendre-en-compte-la-version-dorigine-lors-des-mises-%C3%A0-jour----valid%C3%A9--manuel--official-) : Take into account the original version during updates
```
zamentur: This YEP should be into level 3 no ?

packaging_apps_levels.md
zamentur: +1

packaging_apps_levels.md
zamentur: I was thinking SSO/LDAP to be in level 7. For me, some app with SSO/LDAP could not be High Quality App.

packaging_apps_levels.md
maniackcrudelis: Yep, that looks like a good idea, and a good time to introduce it

packaging_apps_levels.md
maniackcrudelis: But we can't check sso/ldap automatically. That's why I moved those YEP to the level 8.

packaging_apps_levels.md
maniackcrudelis: For me, but maybe I misunderstood this YEP, it's more about apps like nextcloud that needs to be careful about which version it has and which it will reach.
For many apps, most of them even, we just don't care to know from which version and to which we're going to update.

packaging_apps_levels.md
maniackcrudelis: Finally done by https://github.com/YunoHost/package_check/pull/52/commits/5e726b9094301900fa3fb606cd191cf157f59206, all levels are forced to auto, except level 5.

maniackcrudelis:
https://github.com/YunoHost/doc/pull/935 got merged.

This PR is the next in the list.

maniackcrudelis:
No news for more than 20 days.
If no one express a disagreement, I will merge in 3 days.

PR #942, opened 2019-03-03
add a paragraph on cipher setting configuration

rds13:


rds13:
@alexAubin Here the documentation for MR https://github.com/YunoHost/yunohost/pull/640.
I will change the settings name once we have good names.

PR #956, opened 2019-03-19
add post_cert_update hook

Mickael-Martin:
Hi !
Find at this line https://github.com/YunoHost/yunohost/blob/6f89ce071406f6da61c104464f5479fd14455b62/src/yunohost/certificate.py#L814 and works as defined (tested in my fork of openvpn)


alexAubin:
Wow we have some piece of documentation about hooks ô.o

Thanks for the contribution !

PR #978, opened 2019-04-12
Suggestion of two new apps Odoo and BTCpayserver

tarbib:


alexAubin:
(Odoo is already packaged as LibreERP)

PR #994, opened 2019-05-20
Dyndns updates

Jaxom99:
Adding some details about how dynDNS works.
Also mentionning that it's not free/gratis, but paid for and provided by the Project :)


dns_nohost_me.md
```diff
@@ -1,6 +1,6 @@
 # Nohost.me domains
 
-In order to make self-hosting as accessible as possible, YunoHost offers a *free*
+In order to make self-hosting as accessible as possible, the YunoHost Project provides a *free*
```
decentral1se: Perhaps linking to https://github.com/YunoHost/doc/pull/994/files#diff-ab1b71fc907eef6a6ae61cecfc19728aR6 after mentioning "*free*" would also help direct people to the section where it is explained that this is funded by the project.

domains.md
```diff
@@ -3,7 +3,7 @@ Domains, DNS conf and certificate
...
 
-Domains can be managed in the 'Domain' section of the webadmin, or through the `yunohost domain` category of the command line. Each time you add a domain, it is expected that you bought it (or own it) on a domain registrar, so you can manage the [DNS configuration](dns). The exception is the [domains `.nohost.me`, `.noho.st` and `ynh.fr`](/dns_nohost_me) which are free and can be directly integrated with YunoHost.
+Domains can be managed in the 'Domain' section of the webadmin, or through the `yunohost domain` category of the command line. Each time you add a domain, it is expected that you bought it (or own it) on a domain registrar, so you can manage the [DNS configuration](dns). The exception is the [`.nohost.me`, `.noho.st` and `ynh.fr` domains](/dns_nohost_me) which are paid for by the YunoHost Project, and can be directly integrated with YunoHost thanks to an automated dynDNS setup. (To limit costs and abuses, each instance may only have one of these domains setup at any given time).
```
decentral1se: I was wondering about how this was funded. I would recommend it goes under it's own heading? It would be nice to have a separate section making clear the economic considerations of using this facility.

Jaxom99: well, AFAIN, funding is still "in works" within the project : NGO status is pending, and donations are collected through Liberapay, but no official budget/list of spendings has been set up...

decentral1se: Ah, I just meant for the DynDNS stuff, not the entire project. Good to know though :)

alexAubin: Hopefully we should be done with the bank account setup during the next 4 weeks and then we can open a real donation mechanism ;P

frju365:
Ok. Let's merge ! Thanks !

PR #996, opened 2019-06-02
Filter against any match on /app searching

decentral1se:
Closes https://github.com/YunoHost/issues/issues/1347.

I've tested this by copy/pasta'ing the fix into https://yunohost.org/#/apps/edit and playing around with it. I didn't notice any performance issues and it just seems to work ™


decentral1se:
Also see https://github.com/YunoHost/doc/pull/997.

alexAubin:
Thanks !

PR #997, opened 2019-06-02
Add note about how to live test

decentral1se:
It was not obvious to me and I only found out about it by fishing around in the issue tracker for a while. This can save time for people who need to make changes over here. Rendered:

> https://github.com/decentral1se/doc/blob/74a00e054c664256cf833e27763880c6381e419b/README.md


alexAubin:
Supposedly there's an 'Edit' button in the bottom right corner when you're viewing pages ? But anyway it doesn't hurt to add this in the README as well, cheers :wink:

PR #998, opened 2019-06-04
Adding use case YunoHost for non-profits organisations

yalh76:
Not sure :
* it's the good place to place the link to the new page
* if the name I choose for the page is OK

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -31,3 +31,6 @@ YunoHost's documentation has 3 different sections:
    * [Chat rooms](/chat_rooms_en)
    * [Communication](/communication_en)
    * [Support / Help](/help)
+
+* Use example:
+   * [YunoHost for non-profit organisations](/use_case_non-profit_organisations)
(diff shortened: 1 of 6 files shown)


yalh76:
> 
> 
> I think it's a great start. I have many comments but they can come in following PRs where we try to develop this document, no?
> 
> 🚀

It depends if there are a lot of changes to do ^^ would be less work for me if you do your own PRs ^^

decentral1se:
> would be less work for me if you do your own PRs ^^

Yes, that was my idea. But someone has to review those too :)

yalh76:
> 
> 
> Vraiment bien. Par contre, il serait utile de le traduire en fr. :-)

I agree to translate it into French because French is my mother tongue. But in a second time, when @ décentr1se will have done public relations and/or when the page will be almost definitive.

----

Je suis d'accord pour le traduire en français car le français est ma langue maternelle, mais dans un deuxième temps, lorsque @ décentr1se aura effectué ses relations publiques et/ou lorsque la page sera presque définitive.

frju365:
Ok. Let's merge and see if we could improve it ! :-) Thanks !!

PR #1000, opened 2019-06-06
Adding informations on yunohost operations logs

yalh76:


alexAubin:
Uuuuh shouldn't we be talking about `yunohost log list` and `yunohost log display` (and equivalent through Webadmin in Tools > Logs) instead ?

yalh76:
I didn't know these commands, I will change my PR

PR #1012, opened 2019-06-25
Fix links

yalh76:
After installing my own Simone I was able to fix presentation and anchor links


yalh76:
if someone can merge that PR :) thanks in advance ^^

yalh76:
thanks @alexAubin.

@decentral1se: feel free to do PR and let me know when you have added what you want to add. To make a bit of communication around that use example :)

PR #1026, opened 2019-07-18
Fix typos

nicofrand:
BTW the https://yunohost.org/#/use_case_non-profit_organisations_fr/edit does not work it says "Invalid page name" upon submission.


alexAubin:
Thanks !

PR #1050, opened 2019-08-18
Translating from ssh_fr.md

myrrhashad:
- Very specific issue, but on an install-LAN, with lots of devices connected, nmap could help so skim through the results of arp-scan ?
- It feels more user-friendly to indicate when the user can switch over to the post-install after the first login, instead of reading the whole page like I did.


frju365:
Thanks

PR #1051, opened 2019-08-18
Update plug_and_boot_fr.md

myrrhashad:
Markdown was disabled on a line


alexAubin:
Eh I can't reproduce the issue but that sounds okay I guess, apparently this <p> tag doesn't change the rendering anyway

PR #1053, opened 2019-08-18
 Add commands to switch from Bind9 to dnsmasq

myrrhashad:
Just ran into this issue on a fresh install, had to search for the answer : could be worth mentioning directly ?

### DIFF
--- a/install_manually.md
+++ b/install_manually.md
@@ -8,6 +8,14 @@ curl https://install.yunohost.org | bash
 
 <small>*(If `curl` is not installed on your system, you might need to install it with `apt install curl`. Otherwise, if the command does not do anything, you might want to `apt install ca-certificates`)*</small>
 
+<small>*If you get the error message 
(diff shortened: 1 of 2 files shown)


alexAubin:
Hmmm yep we can maybe add the tip about `apt remove bind9` directly [here](https://github.com/YunoHost/install_script/blob/stretch/install_yunohost#L249) I think ? Possibly the same thing about apache ?

I think ideally I'd like to keep the install_manually.md page minimalist and avoid talking about technical edge-case to not scare users that won't run into it, but idk how other people feel about it

myrrhashad:
Yeah, I think if this can be handled by the script then there's no need to change the documentation :)

alexAubin:
@myrrhashad : So I implemented this in https://github.com/YunoHost/install_script/pull/58 , does it look okay to you ?

alexAubin:
The other PR got merged so assuming the issue is fixed naow

PR #1070, opened 2019-09-13
Draft of documentation for groups and permissions

alexAubin:


alexAubin:
Btw if would be nice to have a small paragraph regarding the integration with LDAP in the app configurations. But I don't know enough about it ... ping @Josue-T @zamentur in case you have time for this ...

Josue-T:
> Btw if would be nice to have a small paragraph regarding the integration with LDAP in the app configurations. But I don't know enough about it ... ping @Josue-T @zamentur in case you have time for this ...

Everything is documented [here](https://moulinette.readthedocs.io/en/latest/ldap.html#ldap-integration-in-yunohost-applications)

Gofannon:
1. How can a get the users who belongs to a group? (people from `yolo_crew` for example)
    - what is the output format? "user1;user2"  OR   "user1 user2"    OR    "json"
    - I need this because of how DokuWiki manage admin users at the moment: See https://github.com/YunoHost-Apps/dokuwiki_ynh/blob/testing/scripts/install#L112
2. Is it planned to upgrade the "example_app"?
3. In `you should explicitly delete them in the upgrade script`
    - maybe adding an example: `ynh_app_setting_delete $app $is_public` is relevent?

Gofannon:
Made a github mistake I guess with https://github.com/YunoHost/doc/pull/1070#event-2978070856 :(

PR #1089, opened 2019-09-24
Fix bad link in Yep 1.6

mitra42:
Note - it also  needs someone to edit as to HOW to subscribe to apps@list.yunohost.org - e.g. a link ? - the usual guesses (like sending 'subscribe ' to apps@list.yunohost.org, or going to http://list.yunohost.org don't work


alexAubin:
Hmwell that info about mailing list is old and should be removed, we stopped using those like 2 years ago :/

mitra42:
OK - I removed them as well.

alexAubin:
Thanks !

PR #1095, opened 2019-09-30
Wishlist clean

Plumf:
It's all in the description of the commit, a point of vigilance on my last commit or I deleted wishes that didn't seem to me to be free software or that wasn't intended for a server (like web torrent)

There is still an outstanding question as to what to do with all the services that are no longer maintained from 2013 to 2017?


apps_packaging.md
```diff
@@ -0,0 +1,14 @@
+# Apps packaging in progress
```
alexAubin: Hmmm but those apps are expected to be listed in https://yunohost.org/#/apps, though you gotta choose "All apps" to actually see those ... But I'm really not sure we want to add an additional page like this one to maintain on the long term :/

apps_wishlist.md
```diff
@@ -106,10 +97,8 @@ The following list is a compiled wishlist of applications that would be nice-to-
 - [LocomotiveCMS](https://github.com/locomotivecms/engine)
 - [Logstalgia](http://logstalgia.io)
 - [Loomio](https://www.loomio.org)
-- [MagicMirror²](https://github.com/MichMich/MagicMirror)
 - [MaidSafe](http://maidsafe.net)
 - [Mailpile](https://www.mailpile.is)
-- [Mailtrain](https://mailtrain.org/)
```
alexAubin: You sure about removing those ? (Sorry maybe you already answered that question)

Plumf: No, I'm correcting that.

Plumf: https://github.com/YunoHost/doc/pull/1095/commits/16d5221f0ca78fcf2110db1feba9c9d9b909902a

apps_packaging.md
Plumf: I partly agree with you, that's how I found the links. The proposal results from this [exchange](https://github.com/YunoHost/doc/pull/1056) with @frju365. It was a question for me, and in a longer-term perspective you're right. 

The ultimate solution is to delete these proposals as soon as they are submitted?

Plumf: I totally agree with your analysis! I don't realize what this implies at the technical level.

Plumf:
I chose to keep only the modifications of the wishlist page to avoid unnecessary maintenance.

Plumf:
I deleted all apps that had more than a year of no maintenance. And I added links to the rest even if it makes the page heavier it seems more useful for potential packagers

PR #1119, opened 2019-10-27
Fix app pages

alexAubin:
For instance, https://yunohost.org/#/app_nextcloud_fr is currently broken because of this ...

I guess this was tested in Github and works fine in Github, but Simone's markdown parser works differently and miserably crash because of a missing `href=""` in the anchor ...

Realized while doing this that some stuff were currently a bit overkill / not needed for now ...

Also this thing with link anchors is creating weird behavior because the HTML anchor is already used for the page name ... so e.g. when clicking on a link/anchor inside the app_nextcloud_fr page, the URL in the bar gets changed to `https://yunohost.org/#somelink` instead of something like `https://yunohost.org/#/app_nextcloud_fr#somelink` (but in fact you simply can't have multiple html anchor so ...)

So idk what's the right solution (beside moving to Grav lol idk)

### DIFF
--- a/app_gogs_fr.md
+++ b/app_gogs_fr.md
@@ -10,4 +10,3 @@ Pour utiliser pleinement la puissance de gogs vous devez avoir appréhender ce q
  - [Site Officiel - gogs.io](https://gogs.io)
  - [Dépot applications gogs pour YunoHost](https://github.com/YunoHost-Apps/gogs_ynh)
  - [Site officiel de l'outils git - git-scm.com](https://git-scm.com/)
- - [Trouver de l'aide et poser toutes vos questions : forum.yunohost.org](https://forum.yunohost.org/c/support/apps)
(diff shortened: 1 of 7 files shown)


Plumf:
OMG, it's problematic for a few page, I introduce lots of anchor for apps documentation.

alexAubin:
Merging for now because some people are complaining about the missing pages ...

yalh76:
> 
> 
> For instance, https://yunohost.org/#/app_nextcloud_fr is currently broken because of this ...
> 
> I guess this was tested in Github and works fine in Github, but Simone's markdown parser works differently and miserably crash because of a missing `href=""` in the anchor ...
> 
> Realized while doing this that some stuff were currently a bit overkill / not needed for now ...
> 
> Also this thing with link anchors is creating weird behavior because the HTML anchor is already used for the page name ... so e.g. when clicking on a link/anchor inside the app_nextcloud_fr page, the URL in the bar gets changed to `https://yunohost.org/#somelink` instead of something like `https://yunohost.org/#/app_nextcloud_fr#somelink` (but in fact you simply can't have multiple html anchor so ...)
> 
> So idk what's the right solution (beside moving to Grav lol idk)

To make anchors on a document page you can use as sample what I made for the Table of content in https://github.com/YunoHost/doc/blame/master/use_case_non-profit_organisations_fr.md#L4 or at https://github.com/YunoHost/doc/blame/master/use_case_non-profit_organisations_fr.md#L29

Each title is an anchor, and the id of the anchor is the lowercase title. To find the automated created anchor, sometime you have to install your own Simone ^^

PR #1141, opened 2019-11-03
[enh] add first version of news.md

Psycojoker:
Hello,

After a small discussion with @decentral1se he asked what needed to be done to integrate the news inside Simone. I'm quite out of loop so that might not be the thing you lastly discussed but I came with a prototype of what we've talked about in the past: this create a new page of the doc call "/news", it uses (well, should be, see bellow*) the RSS of the "annoucement" (for now) category using javascript and generate a page for that.

* for now CORS is not enabled on forum.yunohost.org so it doesn't work but the prototype works with the exact same file content

Here is the result for now:

2019-11-03-134818_715x766_scrot

It is not perfect as discourse is quite an idiot and do with stuff like replacing all `` by `<a href="link to image">filename</a>` which can't be easily fixed. And some other weird stuff.

But at least this is a working prototype if we want to build on top of that (if it's the decided direction)

### DIFF
--- a/news.md
+++ b/news.md
@@ -1,0 +1,49 @@
+# YunoHost News
+
+<div id="news">
+</div>
+
+<style>
+.emoji {
+    height: 21px;
+}
+
+blockquote p {
+    font-size: medium;
+}
+blockquote > p {
+    margin-top: 0;
+}
+</style>
+
+<script type="text/template" id="news-template">
+    <article class="news">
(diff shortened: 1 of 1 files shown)


decentral1se:
Once this is merged, I can move forward with the newsletter work. Thanks.

Psycojoker:
Ok, let's merge for now, we'll improve it on the run and it's not easily visible for now.

PR #1144, opened 2019-11-11
Burn or Copy ISO: Change dd instructions from /dev/sdb to /dev/sdz

Wingysam:
It's better for dd to create a file called sdz than it is for it to overwrite someone's second disk (if they have ssd/hdd, that'd be their bulk storage).


alexAubin:
Indeed ! Cheers !

PR #1148, opened 2019-11-28
 [add] non-latin characters in domain names 

tituspijean:
Based on [this thread](https://forum.yunohost.org/t/my-umlaut-broke-my-instance/9666).


frju365:
Thanks.

PR #1152, opened 2019-12-07
Doc for fluxbb, blogotext and webtrees

toitoinebzh:
Doc for fluxbb, blogotext and webtrees


Psycojoker:
Thx! (and sorry for the delay, we are low on manpower :/)

Don't hesitate to send another patch to add the url to the software website later on :)

PR #1154, opened 2019-12-11
Added couchdb to apps wishlist

eauchat:


eauchat:
Maybe one day I'll even be able to package it myself :)

yalh76:
> 
> 
> Maybe one day I'll even be able to package it myself :)

Just try and ask help if needed :)

PR #1166, opened 2020-01-05
[enh] Add a guide to gain access on a broken yunohost

zamentur:
We have no guide to help user to gain access on their server when they lost access.

This document could help them to diagnose and fix their issue.


Psycojoker:
I'm still seeing a todo but as a whole this is already super useful so merging, we'll improve it later if needed.

Thx a lot :heart:

PR #1171, opened 2020-01-21
please dont just allow ssh to root without password

albjeremias:


alexAubin:
Can you elaborate on why you propose this change ... 

This proposed change won't work, because the user 'pi' needs to be disconnected / removed during the (manual) yunohost install on RPi, otherwise we expect people to forget about the existence of the user 'pi', which somewhat leaves a security hole open.

albjeremias:
Sorry. i should i have posted this more as a question.. than as a PR.. 
but leaving the pi user seems a way small security hole, than leaving root without password... is it setting the password in the end? and changing the ssh configs?

alexAubin:
- It's not leaving the setting as "without-password", since it tells : 

>replace `PermitRootLogin without-password` with `PermitRootLogin yes`

- The setting "without-password" does not means that root will have an empty password (supposedly allowing anyone to become root even from a remote location) but instead means that,when authenticating to SSH, you have to use an SSH asymmetric key (which is better than password for various technical reasons) and cannot use the root password, even if it is set

- In this piece of doc, we instead allow root to connect using a password, because that's just way easier for newcomers

- The root password will anyway later be overriden by the admin password chosen during postinsstall

- And the ssh config will also be overwritten by Yunohost's regen-conf mechanism, effectively disabling root password entirely - except from within the local network to still provide an access in case the admin user is broken (e.g. because slapd is down)

zamentur:
So i close this pr

PR #1184, opened 2020-02-15
Fix broken link

Kalkhas:


Kalkhas:
Right, my bad.
It only happens when browsing directly from the repo https://github.com/YunoHost/doc/blob/master/app_nextcloud_fr.md which I guess is  not intended.
Sorry about that.

PR #1188, opened 2020-03-04
I updated the link

MyNameIsTroll:
the link wasn't working


alexAubin:
I don't understand, it works for me... If it doesn't, then that's a bug and probably doesn't affect just this link. Also we don't really want to point to github because the interface is different, and people can't easily contribute, etc.

MyNameIsTroll:
Strange... I tried with Firefox and Edge and I get to a page 404.

github

alexAubin:
That's because you're not supposed to browse the doc from github, you're supposed to browse is through yunohost.org.

For example: https://yunohost.org/#/apps_wishlist

MyNameIsTroll:
Okay, thank you. 
I misunderstood, sorry ^^

PR #1212, opened 2020-03-28
Reuniformize links

alexAubin:
Links are a mess in the doc and it's pretty much not clear for anybody I assume wether to put the lang explicitly or not in the link ... (`/foo` vs `/foo_fr` vs `/foo_en`) ...

So far the confusion was increased by the fact that trying to access to a page not already translated would show you an empty new page ... I tried to address this specific thing in [this Simone branch](https://github.com/YunoHost/Simone/tree/fallback-to-english-if-necessary) which is actually live. You can try this by visiting [this page](https://yunohost.org/#/groups_and_permissions) with your browser being configured as French by default vs. the browser being configured as English by default (before the patch, visiting the page with a French browser would show the "New page" stuff ..)

Anyway, back to the link format issue : I concluded that putting the lang explicitly in the link is problematic, for many reasons which are too complex to detail at 7am ugh

This huge commit is the result of the following small draft epic bash script which we could include to lint the doc. It also identifies dead links (c.f. another fix from earlier).

N.B. : in the current version I still need to fix some links to page that are only available in french (such as `isp_*`, `packaging_apps_*` and `app_nextcloud_*`)

```bash
for FILE in $(ls *.md); 
do 
    # Replace markdown links with full url ... we only need the relative url
    sed -i -E 's@\(https://yunohost.org/#/(\w+)\)@(/\1)@g' $FILE

    # Replace (/foo_fr) to (foo)
    sed -i -E 's@\(\/?((\w|-)+)_(en|fr|es|it|ar|de|oc)\)@(/\1)@g' $FILE

    # Replace href="/foo_fr" to href="foo"
    sed -i -E 's@href="/?((\w|-)+)_(en|fr|es|it|ar|de|oc)"@href="/\1"@g' $FILE;
done

git checkout project_organization.md project_organization_fr.md

# Find all markdown links and generate a list of filename.md:N:linktarget   (with N the line number)
for LINK in $(grep -nr -o -E "\]\(\/?(\w|-)+\)" ./*.md | tr -d ']()/')
do
    PAGE=$(echo $LINK | awk -F: '{print $3}')
    [ -e "$PAGE.md" ] || echo "Issue for this link (page doesn't exist in english?) $LINK"
done

# Find all HTML/href links and generate a list of filename.md:N:linktarget   (with N the line number)
for LINK in $(grep -nr -o -E 'href="\/?(\w|-)+\"' ./*.md | sed -E 's@href="/?@@g' | tr -d '"')
do
    PAGE=$(echo $LINK | awk -F: '{print $3}')
    [ -e "$PAGE.md" ] || echo "Issue for this link (page doesn't exist in english?) $LINK"
done
```


kay0u:
What a huge work! Thanks for that.

Thank make me think that when we tried to release the doc about permissions and group, the website didn't load the MD because there was a missing space between text and source code: https://github.com/YunoHost/doc/commit/ffd82d1c8f91109f64141978b21f4ff6c442eda8.

I don't know if it is a known issue but if we can reproduce this one, we should probably catch it in this new Travis!

alexAubin:
>I don't know if it is a known issue but if we can reproduce this one, we should probably catch it in this new Travis!

Btw about that, was on my way to write some more testing to catch this with some even more obscure regex, but realize that this behavior appears completely random and couldnt identify the pattern ... I guess the real fix would be to find a way to update that damn buggy JS markdown parsing lib [here](https://github.com/YunoHost/Simone/tree/master/_js) (I don't even know if it's markdown.js or marked.js)

PR #1219, opened 2020-03-31
Fix dns path to local ip

Antoine-lb:


Antoine-lb:
I wasn't able to change the language of the interface (changed the language of Firefox but nothing changed). So the french is right, english and spanish should be right but didn't check specifically.

(any hint on how to change the language?)

alexAubin:
Thank you for your contribution,

> (any hint on how to change the language?)

Eh I'm not sure to understand which interface you are talking about ? Github ? YunoHost webadmin ? Yunohost documentation ?

alexAubin:
(I'm merging this but we can keep discussing here, or on the forum or the bugtracker or the chat, whichever you prefer)

Antoine-lb:
Sorry for not being more clear, the question would be:
How can I change the language of the webadmin and documentation easily so I can check what was written and be sure it works.

Of what I understand Yunohost takes the default language, but I didn't manage to change it by modifing Firefox config, what am I doing wrong?

Thanks.

(By chat you mean Kiwi IRC?)

PR #1232, opened 2020-04-06
Groups page - webadmin EN

Jaxom99:
Added (french) screen captures, as a first update. 
Plus some minor text changes.


Jaxom99:
thanks for the quick review @alexAubin :)

I added FR in the same branch/MR, it's simpler for me. Is this still OK ?

PR #1220, opened 2020-03-31
Add git documentation

maniackcrudelis:
Work done.

Can be merged whenever you want.

PR #1234, opened 2020-04-06
git adaptation

Plumf:


packaging_apps_git_fr.md
```diff
@@ -1,72 +1,72 @@
 # Comment utiliser Git pour packager les applications
 
-Git... Notre cher Git bien-aimé, que l'on peut aussi décrire comme "Goddamn Idiotic Truckload of sh*t", selon Linus.  
+Git... Notre cher Git bien-aimé, que l'on peut aussi décrire comme "Goddamn Idiotic Truckload of sh*t"  ( P *t**n de camion de m*rde), selon Linus.
```
maniackcrudelis, line 3: That's a literal translation, and as I was quoting Linus Torvalds, so I kept the original version.
To translate that would looks like... "Un stupide putain gros tas de merde" Does it really worth it ?

packaging_apps_git_fr.md
```diff
@@ -1,72 +1,72 @@
...
 Si vous ne connaissez pas encore Git, soyez sûr que vous serez bientôt d'accord avec cette description.
 
-YunoHost et toutes nos applications sont sur la forge Git GitHub. Ce qui veut dire que si vous voulez travailler sur une application, tôt ou tard vous allez devoir faire face à Git.
```
maniackcrudelis, line 6: I was wondering why you did have to add back the double space at the end of the lines.
But actually, that's the opposite !

Those double space are here in purpose, so it adds a new line.

packaging_apps_git_fr.md
Plumf: yes, Seems to me it keeps the spirit of the quote.

maniackcrudelis: I guess the original quote is not so obvious for non English speakers.
So, if you think it really worth it, why not.
But, do not censor more than the original quote. A censured quote from Linus isn't anymore from him !

packaging_apps_git_fr.md
Plumf: I didn't add double lines, I was wondering why they were there, but I didn't touch them. For information I use Atom for writing, maybe this software change this ?

maniackcrudelis: Probably the common auto correction of removing trailing spaces at the end of lines.
Better to keep them.

maniackcrudelis:
Planning to merge #1220 tomorrow, if you have some time, could you finish that PR, so we can merge into #1220 and go on.

Plumf:
So it's ok for me, thank.

maniackcrudelis:
Huh no it's not ok, you didn't fix that PR.
What do we do with the translation of Linus quote ?

And the double spaces at the end of lines should be put back.

Plumf:
Sorry, I'm ok for Linus quote I fixed that, but the double spaces are deleted by my notepad as soon as I save the document it deletes them directly...

maniackcrudelis:
You're going to need to fix that, you can write markdown without spaces at the end of your lines.

packaging_apps_git_fr.md
Plumf: I trie to mchange parameters

PR #1277, opened 2020-04-29
[baikal] clarify users management

nqb:


nqb:
I think we can also drop "Creating a new user" section as its useless.

Any opinion on that @kay0u and @Salamandar ?

app_baikal.md
```diff
@@ -24,10 +24,16 @@ Baïkal is a server for calendars and address books, which uses the CalDav and C
 To configure the app, go to the address: `sub.domain.tld/admin` or `domain.tld/baikal/admin`.
 The username to specify is `admin`, followed by the specific password you chose when installing Baïkal. The password can contain any special characters.
 
+### LDAP authentication
+
+By default, Baikal is configured to lookup users in YunoHost's LDAP
+directory. YunoHost's users will appear under `User and ressources` menu after their
+first successful authentication.
```
comradekingu, line 31:
```suggestion
first authentication.
```

app_baikal.md
```diff
@@ -24,10 +24,16 @@ Baïkal is a server for calendars and address books, which uses the CalDav and C
 To configure the app, go to the address: `sub.domain.tld/admin` or `domain.tld/baikal/admin`.
 The username to specify is `admin`, followed by the specific password you chose when installing Baïkal. The password can contain any special characters.
 
+### LDAP authentication
+
+By default, Baikal is configured to lookup users in YunoHost's LDAP
```
comradekingu, line 29:
```suggestion
By default, Baikal is configured to look for users in YunoHost's LDAP
```

nqb:
Thanks for your suggestions @comradekingu.

Salamandar:
> I think we can also drop "Creating a new user" section as its useless.
> 
> Any opinion on that @kay0u and @Salamandar ?

Hum… I guess… I mean, changing the webDAV authentication type would not use LDAP anymore ? Keeping the existing users from logging in ?

Salamandar:
I confirm myself : I changed the auth setting on my server and I couldn't update my address books anymore. Changing the authentication method back to LDAP resolved the issue.

I guess, to prevent any issue on the users' side, this section should indeed be deleted.

nqb:
Hi @Salamandar,

From what I see in git log, it was certainly an old installation instruction when Baikal was not plugged on YunoHost's LDAP. 

The only reason I see to create local users to Baikal could be to give access to users that didn't exist in YunoHost.

Salamandar:
@nqb The issue is that it breaks Baikal for existing Yunohost users. It's either local OR LDAP, not both. Unfortunately.

nqb:
PR updated, if everybody is fine with that state, let's go merge.

PR #1302, opened 2020-05-16
create try_ca.md

xaloc33:
Create and add catalan translation to try file

I think that I have not broken any of the syntax I've tried to be careful


alexAubin:
Thanks ! :heart:

PR #1318, opened 2020-05-24
WIP: Updated french text in english doc

vrubiolo:
This fixes a potential leftover from the original translation.

PR #1319, opened 2020-05-24
Updated french text in english doc

vrubiolo:
This fixes a potential leftover from the original translation.

### DIFF
--- a/app_wallabag2.md
+++ b/app_wallabag2.md
@@ -29,6 +29,6 @@ In addition to Wallabag core features, the following are made available with thi
 
 ### Upgrade from v1.x
 
-No automatic upgrade process is available. You need a manual (but simple) migration from [Wallabag v1](/app_wallabag).
+No automatic upgrade process is available. You need a manual (but simple) migration from Wallabag v1.
 Please take a look at the [official documentation](https://doc.wallabag.org/en/user/import/wallabagv1.html).
 
--- a/app_wallabag2_fr.md
+++ b/app_wallabag2_fr.md
@@ -28,6 +28,6 @@ En plus des fonctionnalités principales de Wallabag, ce paquet propose égaleme
 
 ### Mettre à niveau depuis la v1.x
 
-La mise à niveau depuis le paquet Yunohost de [Wallabag v1](/app_wallabag) demande une opération manuelle, c'est pourquoi un nouveau paquet est fourni. Pour le processus de migration, merci de vous référer à [la documentation officielle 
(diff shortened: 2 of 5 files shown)


vrubiolo:
This is a revival of #1318 with a properly setup repo (no more a fork of a fork). Let's see how this goes...

vrubiolo:
Alright, we are in better shape now, the 2 commits above fixed 2 of the 3 missing tests reported by Travis. 

Some notes about them:
   * for the Wallabag fix, it looks like the v1 app is not available anymore so I chose to remove the link entirely (as we are already pointing at the Wallabag doc itself).
   * for `groups_and_permissions_fr.md`, the page initially pointed at `https://yunohost.org/#/groups_and_permissions_en` which was reported as an error by the script (rightfully, as there is no such file) but the link kinda worked and ended up at the bottom of the English page. The updated link correctly points at the top of the english page and fixes the error reported by CI. Just wanted to confirm this is OK w/ you.

For the 3rd error (unreferenced pages), I don't know what to do here so I have left it as is.

/ready-for-review-for-me

groups_and_permissions_fr.md
```diff
@@ -173,4 +173,4 @@ La webadmin émettra un avertissement si vous définissez une permission qui est
 Notes pour les packageurs d'applications
 ------------------------
 
-[Voir la version anglaise de cette page](groups_and_permissions_en).
+[Voir la version anglaise de cette page](groups_and_permissions).
```
alexAubin, line 176: Eeeeh unfortunately that one is intentional because it's meant to point to the english version explicitly ... but I understand that the tests report it (I wrote them ;P) . I think in fact it's bad designed to have the english version of this page include a huge section only meant for packagers ... we should create a dedicated page for the maintainer infos which would solve the issue...

vrubiolo: @alexAubin : no worries, and thanks for the explanation. I thus understand I should leave this file unchanged, is that right?
As for the better/full solution, you are advising we split out the packager-only instructions out of the english page and move it to a dedicated page, then translated in both English and French, is that it?

alexAubin: Yes, but no need to translate the page .... that's a documentation meant for packagers and most of them are used to deal with english except of somebody really is motivated ... except if somebody really is motivated about it :/

vrubiolo: Ok, thanks for the feedback, I have reverted the change on that file and squashed everything so that it looks cleaner.

vrubiolo:
Ping! I think we are in good shape now :smile:

alexAubin:
Hey there ! Thanks for the ping ! Yolomergin !

PR #1329, opened 2020-06-12
Apps hooks translation into english

ericgaspar:
- *Add english translation*
- *Small typos for the french*

### DIFF
--- a/packaging_apps_hooks.md
+++ b/packaging_apps_hooks.md
@@ -1,1 +1,191 @@
-Unfortunately, this page only exists [in french here](packaging_apps_hooks_fr) for now.
+# The use of YunoHost hooks
+
+Hooks allow you to trigger a script when an action is performed by the system.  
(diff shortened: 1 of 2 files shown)


packaging_apps_hooks.md
```diff
@@ -1 +1,191 @@
...
+cp hook_regen_conf /usr/share/yunohost/hooks/conf_regen/50-SERVICE_$app
+```
+
+> When removing the application, this hook must be removed manually.
+
+##### Building `conf_regen` hook script
+
+`conf_regen` hook is called two times, a first time after analysis of the configuration and before any modification of the files, then a second time after applying the modifications, if there has been modifications.
+
+`conf_regen` hook script should look like this:
```
alexAubin, line 159: I'll try to add a simpler example for this later, as usually the use case for this is to patch an existing config file and therefore only the "pre" step matter ... and there's some trick to edit the default conf ... c.f. for example this hook : https://github.com/YunoHost-Apps/my_webapp_ynh/blob/testing/conf/ssh_regenconf_hook

PR #1361, opened 2020-07-29
Update hardware.md

tido-:
I tried to run it on 512 MB device. Installation of NC or InvoiceNinja simply break and will never run. To go with 512, will simply end in a frustration for the user.

### DIFF
--- a/hardware.md
+++ b/hardware.md
@@ -10,5 +10,5 @@ Corresponding installation guides can be found on [this page](/install).
 ### Minimum requirements
 
 * 500 MHz CPU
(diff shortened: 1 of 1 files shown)


zamentur:
Don't know what to do with that. I mean it could be possible nextcloud install fail for this reason, but the diagnostic ask for setting a swapfile. If you have swapfile you should be able to install nextcloud. And nextcloud+ynh run on 512MB.

So i suggest to close this PR. Note the hardware page has been deleted. 
We could imagine warning on the prerequisite part or on the diagnostic part about swapfile ?

zamentur:
Done 84aaacf6e2fe725fc17c6c23b169563921798f1d

PR #1435, opened 2020-11-12
add Mobilizon to wishlish

RayBB:
Made by the same folks who build PeerTube!
Learn more here: https://joinmobilizon.org/en/


tituspijean:
Mobilizon is already packaged: https://github.com/YunoHost-Apps/mobilizon_ynh

PR #1449, opened 2020-11-26
Add Twake in the list

dsmrs:
[Twake](https://twake.app/fr/home) All of the organizational tools you need to make your projects succeed brought together on one platform.

A mix project management+chat+agenda+docs = MS Team but opensource (GPLv3)


Plumf (approved):

frju365 (approved):

frju365:
Thanks !! :smiley: :+1:

tituspijean:
One note though, Twake only relies on Docker for self-hosting, which is not the standard way for YunoHost and would make it difficult to package. 😟

frju365:
yeah :/ Just saw this. Well, can be still docker installation as we did previously with Collabora/Onlyoffice during a period.

PR #1491, opened 2021-01-06
Advertise the use of yunohost.local for install @ homes

alexAubin:
fix https://github.com/YunoHost/issues/issues/869

- Since 4.1, people should now be able to reach their server using yunohost.local AND run the posinstall (previously the postinstall would sort of crash if done from yunohost.local)
- We should advertise this in the doc instead of the damn technical step of "lol now have fun and find your local IP"
- The install doc for RPi / ARM board was in fact buggy because recommended to log on SSH despite the fact that you don't need it (even if you don't use yunohost.local)

**Before merging this, we should make sure that all the images are upgraded to 4.1 ...**


finding_the_local_ip_fr.md
```diff
@@ -0,0 +1,19 @@
+# Trouver l'IP locale de son serveur
+
+Votre serveur devrait typiquement être accessible (depuis son réseau local) avec le domaine `yunohost.local`. Si pour une raison cela ne fonctionne pas, il vous faut peut-être trouver l'IP locale de votre serveur.
+
+L'IP locale d'une machine est utilisée pour y faire référence à l'intérieur d'un réseau local (typiquement le réseau dans une maison) où plusieurs appareils se connectent à un même routeur (votre box internet). Une adresse IP locale ressemble généralement à `192.168.x.y` (ou parfois `10.0.x.y`)
+
+L'une de ces astuces devrait permettre de trouver l'IP locale de votre serveur:
+
+- Ytilisez l'interface de votre box internet pour lister les machines connectées, ou regarder les logs ;
```
nqb, line 9: Utilisez

ericgaspar (approved):

MrMorals:
So happy with this! I don't know the technical details of this well enough to dare change the documentation in a big way. But looking forward to being able to subtract a an hour or two from my next Yunohost installation process :smile:

PR #1553, opened 2021-02-14
Add Signal Proxy

JanaKeyson:
https://signal.org/blog/help-iran-reconnect/


alexAubin:
Cheers !

Merging before the big migration to Grav in a few second

PR #1583, opened 2021-02-22
fix link to tool 'send_to_dev_ci.sh'

Gofannon:


alexAubin:
(Note that this workflow is obsolete and most probably not working anymore, c.f. the note at the beginning of the page, nowadays we use the comment-!testme-on-the-PR mechanism)

Gofannon:
> (Note that this workflow is obsolete and most probably not working anymore, c.f. the note at the beginning of the page, nowadays we use the comment-!testme-on-the-PR mechanism)

Thanks for the tip! Will try.
I wasn't aware of this as I have been away for some time. (and I didn't read the whole page obviously - what a (nice) change for the doc.

I have some questions as I'm here:
1. Can `package_check` still be run on my computer? (I see a switch to LXD) My goal is to test before pushing the code to a new PR to fix things before pushing on the public repo.
2. Do you, I mean the "YunoHost CI", still need some raspberry pi to run `package_check` on it? I used to land a pi but it blew up and I have never been able to reinstall pcheck on it... Maybye with the new version, I could spend time on it if it's worth it for the CI?

alexAubin:
>Can package_check still be run on my computer? (I see a switch to LXD) My goal is to test before pushing the code to a new PR to fix things before pushing on the public repo.

Yes I believe it's in fact *safer* now because the code doesn't require sudo anymore (though the LXC is launched as privileged) and also it doesnt savagely tweaks iptables rules of the host machine, everything in handled in a cleaner way using LXD directly (though LXD can be sometimes troublesome to get running but anyway)

>Do you, I mean the "YunoHost CI", still need some raspberry pi to run package_check on it? I used to land a pi but it blew up and I have never been able to reinstall pcheck on it... Maybye with the new version, I could spend time on it if it's worth it for the CI?

To be discussed with Kayou, for now we dropped the "submit jobs to a pool of rpi-workers" because that was overly complexed and not reliable. I think the main plan is to go for a bigger board like a RPi4 with 4+ GB or RAM to have a simpler, more reliable CI. Maybe we can have a system at some point to send jobs to workers but that requires development.

Gofannon:
Thanks for the input Alex!

PR #1636, opened 2021-03-27
Remove escape character

ericgaspar:
- Remove escape character for NodeJS helper

I'm not sure if this is the correct way to correct it as this document seems to be auto-generated ...


tituspijean:
Yup indeed, it is [generated](https://yunohost.org/en/contribute/write_documentation#special-pages) from the comments preceding the [helper definition](https://github.com/YunoHost/yunohost/blob/adc83b4c9c2c30e9ef75f3609c538b646f91f1db/data/helpers.d/nodejs#L29). The text over there looks alright though, so I guess it is a bug with the generation script.

PR #1643, opened 2021-04-05
Update apps_wishlist.md

ericgaspar:
- *Add Photostructure and PicApport to the wishlist*


alexAubin:
Sounds like Photostructure ain't free software tho https://github.com/photostructure/photostructure-for-servers/blob/main/LICENSE.md :/

PR #1679, opened 2021-05-14
Note about smtp.relay.password and ASCII only

jedie:
See also: https://github.com/YunoHost/issues/issues/1800


zamentur:
Thanks for your contribution

PR #1683, opened 2021-05-21
Update security.md to use ssh port setting

tituspijean:


tituspijean:
Let's maybe update the SSH keys doc, too?

noo1ang8:
Be careful, there is a typo : to give a value to `yunohost`, the option is `-v`.
And the French explanation has been copy/pasted over from the English text…

tituspijean:
Fixed. To be squashed-and-merged.

PR #1684, opened 2021-05-23
packaging_app_helpers language reworked

comradekingu:
Makes it easier, but I will have made some mistakes, so unfortunately it all needs to be reviewed carefully.
There are a bunch of missing ending dots, but that is not the problem we have.

---

**Didn't understand or problematic**:

. And so for    'pm.start_servers',      

fi Requires YunoHost version 4.1.2 or higher.`     

psql  ## is this PostgreSQL, or the front-end software for it?

global settings ## I never like this. Program-wide or system-wide settings?
global variables   ## same thing.

call this script in install and restore scripts.  ## in "install and restore scripts" section?

String that will replace matches    ##  Exact match to replace?

Underscorify the string (replace - and . by _)   ## Wouldn't that underscorify it?

uncompress and patch the source from app     ##  Can't remember

Instead of true, SOURCE_IN_SUBDIR could be the number of sub directories to remove.        ## Explain?

**Suspect**:

sha256sum, md5sum…). Default: sha256    ## not sha256sum?
jessie, stretch, …)   ## not Jessie, and Stretch?


pages/04.contribute/04.packaging_apps/11.helpers/packaging_apps_helpers.md
```diff
@@ -1418,57 +1418,57 @@ Requires YunoHost version 2.7.12 or higher.
...
-- `-A`, `--additional_urls=`: (optional) List of additional URL for which access will be allowed/forbidden
-- `-h`, `--auth_header=`: (optional) Define for the URL of this permission, if SSOwat pass the authentication header to the application. Default is true
-- `-a`, `--allowed=`: (optional) A list of group/user to allow for the permission
-- `-l`, `--label=`: (optional) Define a name for the permission. This label will be shown on the SSO and in the admin. Default is "APP_LABEL (permission name)".
-- `-t`, `--show_tile=`: (optional) Define if a tile will be shown in the SSO. If yes the name of the tile will be the 'label' parameter. Defaults to false for the permission different than 'main'.
-- `-P`, `--protected=`: (optional) Define if this permission is protected. If it is protected the administrator won't be able to add or remove the visitors group of this permission. Defaults to 'false'.
+- `-A`, `--additional_urls=`: (optional) List of additional URLs for which access will be allowed/forbidden
+- `-h`, `--auth_header=`: (optional) Define for the URL carrying this permission, if SSOwat should pass the authentication header to the app. The default is "true".
```
comradekingu, line 1422:
```suggestion
- `-h`, `--auth_header=`: (optional) Define for the URL carrying this permission, if SSOwat should pass the authentication header to the application. The default is "true".
```

alexAubin:
Uh okay but as stated at the top, that file is in fact auto-generated from [this script](https://github.com/YunoHost/yunohost/blob/dev/doc/generate_helper_doc.py) that parses the comments in https://github.com/YunoHost/yunohost/tree/dev/data/helpers.d ...

tituspijean:
As stated by Aleks, you can propose changes in the comments of the linked helpers definition files.

PR #1690, opened 2021-05-30
Fix some typo in english security doc

rds13:


alexAubin:
Eeeeh, "it's not a typo, it's a feature" ? In grav's markdown syntax, those `!!!` are here to create warning/info alert boxes

rds13:
I noticed that it does not render well at least on Github and that the french version used an alternate syntax to highlight the statements.

tituspijean:
We use Grav to render the documentation, GitHub not rendering it is not an issue to us. English version was the most up-to-date, let's keep it that way.

PR #1710, opened 2021-06-28
Added Prosody IM

edo0:
Prosody is a server for Jabber/XMPP written in Lua. It aims to be easy 
to use and light on resources. For developers, it aims to give a 
flexible system on which to rapidly develop added functionality or 
rapidly prototype new protocols.


edo0:
Wrong git address, will open a new pull request

PR #1711, opened 2021-06-28
Added Prosody XMPP

edo0:
Prosody is a modern XMPP communication server. It aims to be easy to set up and configure, and efficient with system resources. Additionally, for developers it aims to be easy to extend and give a flexible system on which to rapidly develop added functionality, or prototype new protocols.


tituspijean (approved):

ericgaspar:
Not sure about the pertinence of putting Prosody in the wishlist. YunoHost uses Metronome over Prosody, it is part of the core.

See this issue: https://github.com/YunoHost/issues/issues/67

alexAubin:
Yeah ... though maybe one day we'd make Metronome optional and maybe you could be able to install Prosody (assuming all the corresponding regen conf hook would also be able to handle Prosody), but that's a lot of 'maybe' and 'if's

PR #1717, opened 2021-07-11
Add smartypants extension to enhance typography

tituspijean:
cf. https://github.com/YunoHost/doc/pull/1716

Adding `smartypants` extension beautifies and respects typography laws by using the curved apostrophes and quotation marks, and even offers to display French guillemets. Documentation has been updated accordingly.

- [ ] Backend task @tituspijean: install said extension.

### DIFF
--- a/config/plugins/smartypants.yaml
+++ b/config/plugins/smartypants.yaml
@@ -1,0 +1,10 @@
+enabled: true
+twig_filter: true
+enabled_in_admin: false
+process_title: true
+process_content: true
+options: qDewgf
+dq_open:  “
+dq_close: ”
+sq_open:  ‘
+sq_close: ’
--- a/pages/04.contribute/02.write_documentation/write_documentation.fr.md
+++ b/pages/04.contribute/02.write_documentation/write_documentation.fr.md
@@ -27,14 +27,17 @@ La structure du dépôt est décrite ici:
 +-- config
    +-- site.yaml
    +-- system.yaml
+   +-- plugins
+       # Contient les fichiers de configuration des plugins.
+       # Attention, ne partagez pas d'information sensible ou des identifiants.
    +-- themes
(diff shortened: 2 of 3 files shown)


zamentur:
I don't see how it doesn't apply to other content than fr ?

tituspijean:
It indeed does not discriminate between languages for now. However, per Wikipedia, both [apostrophes](https://en.wikipedia.org/wiki/Apostrophe#Typographic_form) and [quotation marks](https://en.wikipedia.org/wiki/Quotation_mark#Summary_table) in English can be used with their curved, typographic version.

This PR does not specifically target French, even though the issue arose from the French translation. From the second link I see we may have issues with the German version. I could have a look into it, could be a simple `if` to apply somewhere in the code.

Psycojoker:
idk if this correctly take care of avoiding reformating code snippets like in `<pre>` of `<codeblock>` tag, otherwise that could break a lot of command line that people cold copy paste (already experienced that)

tbh I find people forcing for The Right French Typography so annoying that I don't want to hear about it anymore but that's not a real argument.

tituspijean:
> idk if this correctly take care of avoiding reformating code snippets like in `<pre>` or `<codeblock>` tag, otherwise that could break a lot of command line that people cold copy paste (already experienced that)

You can check it out on my test server (`ynh.pijean.ovh/fr/packaging_apps_arguments_format`). Code blocks are excluded (as stated in the doc of the extension).

> tbh I find people forcing for The Right French Typography so annoying that I don't want to hear about it anymore but that's not a real argument.

I hear you. I am merely proposing a lazy way of handling it, while maintaining our habits of using `'` and `"`.

tituspijean:
The plugin clashes with the Shortcode-UI tabs extension. Do *not* merge.

PR #1739, opened 2021-07-27
Add note to install dialog

gisforgabriel:
Not all providers have dialog in their Debian 10 image. There is an error message shown, but it is not an intuitive diagnosis for all users.

### DIFF
--- a/pages/01.administrate/05.install/install.md
+++ b/pages/01.administrate/05.install/install.md
@@ -476,6 +476,8 @@ curl https://install.yunohost.org | bash
 
 !!! **Note for advanced users concerned with the `curl|bash` approach:** consider reading ["Is curl|bash insecure?"](https://sandstorm.io/news/2015-09-24-is-curl-bash-insecure-pgp-verified-install) on Sandstom's blog, and possibly [this discussion on Hacker News](https://news.ycombinator.com/item?id=12766350&noprocess).
 
+!!! If `dialog` is not installed on your system, you will soon be shown an empty prompt which hangs. Exit the script with <kbd>Ctrl</kbd> + <kbd>C</kbd> and install it with `apt install dialog`.
+
 {% endif %}
 
 


tituspijean:
That's weird, the install script already takes care of requesting `dialog`, if the script is not called in `AUTOMODE`. I am wondering if you actually found a bug.

gisforgabriel:
@tituspijean Ah sorry I just assumed that it was a small thing missing from the docs and didn't think to see whether it's actually an issue with the code.

In terms of debugging - I'm not sure how to verify exactly what image my provider is using to build my VPS. It's just labelled `Debian 10 64bit`. It was a clean install, with literally just an `apt update` and `apt upgrade` before curl-ing the installation script. Is that enough to go on? 🙈

tituspijean:
If you open `install.yunohost.org`, you can see the install script. The `install_script_dependencies` command directly quotes `dialog` and should have it installed.

In any case, a PR has to be done, either yours updating the documentation, or one to fix the install script. Let's keep yours open while we have a look into it. 😇

gisforgabriel:
This is what my initial installation attempt looked like, for reference:

```
root@myservername:~# curl https://install.yunohost.org | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 22294  100 22294    0     0  72383      0 --:--:-- --:--:-- --:--:-- 72383
[INFO] Running upgrade_system
debconf: unable to initialize frontend: Dialog
debconf: (No usable dialog-like program is installed, so the dialog based fronte                                                                                            nd cannot be used. at /usr/share/perl5/Debconf/FrontEnd/Dialog.pm line 76.)
debconf: falling back to frontend: Readline
Installing packages
-------------------

..100%


^C
```

But... just realised that my installation of YunoHost is having trouble calling to external resources, for example the catalog json:

```
Error: "500" Internal Server Error

Action: "GET" /yunohost/api/apps/catalog?full&with_categories&locale=en

Error message:
Unable to download the default app catalog: Invalid URL https://app.yunohost.org/default/v2/apps.json (does this site exists?)
```

This might be why it couldn't request a download of dialog (the others dependencies were probably already in place)?

I think this is actually a bug at my end and I should move it to issue tracker, but it might be worth putting an early test into the install script to see if external resources can be called? (I don't know enough to know if this makes sense - just an idea 😄).

tituspijean:
I looks like the very first command `upgrade_system` already needs `dialog`, while the code for installing this dependency is on the second one, `install_script_dependencies`. I have pinged the devs to have look here. ;)

alexAubin:
Fixed in install script via https://github.com/YunoHost/install_script/commit/6103c75c947b7f4f31f457aef915d0ea73d46fd8

PR #1749, opened 2021-08-12
Fix typos in Administration

sagessylu:
- Fix typos in some of the French administration part (from beginning to the end of overview)
- Standardize things such as "webadmin", "votre.domaine.tld"...


tituspijean (approved):
Thanks for your work!

We had a recent debate about the French quotation marks, and it seems we would prefer to keep the bad straight English ones `"`. However, the automated replacement plugin does not work yet, so I am inclined to allow them here. We can always replace them later.

PR #1758, opened 2021-08-23
Update apps_wishlist.md

rungeard:
Add Quizzes to the wishlist


ericgaspar:
Looks like this app has not been maintained for 8 years ...

rungeard:
I agree with you. But it's very difficult to find a project that allow you to create and serve quiz online with the same features as this one (once installed everything could be done in a web interface, no IT skilled is required, which is I think very interesting for YunoHost Users). 
I have also seen https://github.com/tecnickcom/tcexam which seems pretty well, especially because it's translated in many languages, but the license is more ambiguous and it seems very complex, doing much more that what a single teacher may be looking at.

ericgaspar (approved):

alexAubin:
@rungeard possibly framaforms (recently renamed Yakforms) may fit your needs (not super sure about this, i don't know exactly the features and need)

rungeard:
Thanks for your suggestion @alexAubin !

I think that Yakforms is more thought in a polling/survey way (no right or wrong answer, only points of view).

I was rather looking for an app designed for educational purposes to edit, distribute and correct online quizzes for students. The objective being to assign a grade to each student and let him know where he stands in a class. 

I think I'll dig into Quizzes in more detail because it seems to me a good base. If necessary I will make some pull requests after testing it in a development environment to make sure that the dependencies are still maintained, and why not to add i18n support.

ericgaspar:
Quizzes as been packaged https://github.com/YunoHost-Apps/quizzes_ynh

PR #1774, opened 2021-09-05
? to !, as it isn't a question

dhuschde:
Changes the German term for Host it yourself.
It isn't a question, so it shouldn't use ?


tituspijean:
I am a bit confused, the original sentence is indeed a question: https://yunohost.org/en/howtohostyourself
I do not know German, maybe you can rephrase the whole title to make it into a question, if needed?

...But now I realize that it can indeed be understood as a declarative sentence. Like, "Here is how to host yourself."

dhuschde:
I think, as a question, it would'n sound as good...

frju365 (approved):
I agree with you. it seems to be better with "!"

PR #1789, opened 2021-09-25
Add Cloudflared to the wishlist (not free software)

martinkeat:
Cloudflared also allows automatic DNS updates with argo tunnels to create proper DNS entries. Unfortunately I lack the experience in order to properly package it and integrate into yunohost.

### DIFF
--- a/pages/02.applications/04.wishlist/apps_wishlist.md
+++ b/pages/02.applications/04.wishlist/apps_wishlist.md
@@ -42,6 +42,7 @@ You can [contribute to this list by adding something you'd like to be packaged](
 | [CheckUp](https://sourcegraph.github.io/checkup) |  | [Upstream](https://github.com/sourcegraph/checkup) |  |
 | [Citadel-suite](https://www.citadel.org) | Groupware platform |  |  |
 | [CloudTube](https://tube.cadence.moe/) | CloudTube front-end for YouTube | [Upstream](https://git.sr.ht/~cadence/cloudtube) |  |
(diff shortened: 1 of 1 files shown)


alexAubin:
(Looks like this is not free software...)

martinkeat:
> (Looks like this is not free software...)

You can create an argo tunnel without actually signing up to their tunnel service - and using their free dns service. You simply create the tunnel using the "cloudflared" process, then create a cname record with <tunnelidnumber>.cfargotunnel.com 

See here: https://www.reddit.com/r/Ibracorp/comments/orwl11/cloudflare_how_to_set_up_cloudflare_argo_tunnel/

That guide is about setting it up for free to unraid. It would be similar process for dns on yuno (but obviously tweaked so it works with it).

The beauty of an argo tunnel is no requirement to open ports on your router which is great for those setting up home servers and if setup on yunohost to install automatically, it's going to make life easier when it comes to DNS and will afford security that the person wouldn't otherwise have.

martinkeat:
So yeah, it's free to setup argo tunnels.

martinkeat:
Cloudflared is also free.

alexAubin:
What I meant was free as in freedom, not as in free beer : their LICENSE doesn't seem to be free software : https://github.com/cloudflare/cloudflared/blob/master/LICENSE and the Yunohost project will only include free software (or possibly not-totally-free-software ;)) in the official app catalog

martinkeat:
1.1. ...."You may also use the Service to create derivative works which are exclusively compatible with any CloudFlare product service and no other product or service."

Considering it is only used to interface with argo tunnels it's addition would be for the purposes of making yuno compatible with cloudflare argo tunnels so we're golden! :-) Besides most of the code that they used for it is open source itself.

alexAubin:
I'm not arguing wether it's useful or not, I'm just stating the fact that this looks like a custom license which apriori (i'm no license expert) looks like it's not part of what is called "free software" (which is different from being "open source"), and therefore has implications in terms of integration in Yunohost's app catalog if it were to be packaged

PR #1792, opened 2021-09-28
Update commands

theQureshi:
The upgrade commands now work by adding double dashes before "system" and "apps"

PR #1793, opened 2021-09-28
Update upgrade.md

T3hlo:
Added "--" to options to reflect current syntax.


tituspijean:
Duplicate of litterally today's PR: https://github.com/YunoHost/doc/pull/1792.

That is not current syntax anymore, but I guess we are in a chicken and egg situation, because to upgrade you need the old syntax...

PR #1812, opened 2021-11-02
Zen offer fixed IP as standard

bailtree:
Was their customer, can confirm.


tituspijean:
Thanks for your contribution! 🚀

PR #1823, opened 2021-11-20
Add CyberChef to the Apps Wishlist

AlexioTROM:
Add CyberChef - https://gchq.github.io/CyberChef/ - a simple, intuitive web app for carrying out all manner of "cyber" operations within a web browser. This tool is designed to enable both technical and non-technical analysts to manipulate data in complex ways without having to deal with complex tools or algorithms. It was conceived, designed, built and incrementally improved by an analyst in their 10% innovation time over several years.


tituspijean:
And I love it is being created and maintained by GCHQ.

PR #1856, opened 2021-12-22
some new spanish translations and some simplifications on already translated documents, leveriing the original technial and knowledge requeriments to the french version

juanalbglz:


alexAubin:
Thanks for all the work!

PR #1859, opened 2021-12-27
Added Inventaire

xplosionmind:


tituspijean (approved):
Uh, I had a strong déjà vu with this one, but nope it is not packaged. 😅

alexAubin:
@tituspijean : maybe you're thinking about pyinventory :P

PR #1862, opened 2021-12-28
Add Open Streaming Platform to the wishlist

cnred-dev:
The Open Streaming Platform - https://openstreamingplatform.com/ - was designed as a self-hosted alternative to services such as Twitch.tv, Ustream.tv, Mixer, and Youtube Live' and is a Live Streaming App. Use the demo server to test OSP at https://demo.openstreamingplatform.com/


ericgaspar:
Open Streaming Platform is already packaged (but it is not working at the moment...) https://github.com/YunoHost-Apps/osp_ynh

PR #1875, opened 2022-01-12
Update and redesign with more precision for English documentation

McFlyPartages:
EN Redesign of the documentation by adding more precision

FR Refonte de la documentation en ajoutant plus de précision


ericgaspar:
superseded by #2069

PR #1884, opened 2022-01-22
Update apps_wishlist.md

Thatoo:
Add Raspap


tituspijean:
We have `hotspot_ynh`, but this one ships seemingly more features.

PR #1893, opened 2022-02-01
Adding mail-in-a-box to wishlist

DokStook:


alexAubin:
Adding mailinabox to Yunohost would be very odd considering Yunohost already ships its own mail stack, so both are gonna conflict ...

DokStook:
Alright, didn't consider that

PR #1894, opened 2022-02-02
Added Excalidraw

xplosionmind:


tituspijean:
Excalidraw is already packaged: https://github.com/YunoHost/apps/blob/master/apps.json#L816-L823

PR #1918, opened 2022-03-22
Add 'selfoss' in apps whishlist

Binnette:
selfoss: web based rss reader and aggregator https://selfoss.aditu.de/


ericgaspar:
It is already in catalog: https://github.com/YunoHost-Apps/selfoss_ynh

PR #1993, opened 2022-05-09
Update apps_wishlist.md

scriptor-pro:
polr - un chouette racourcisseur d'url


ericgaspar:
polr is already packaged (may not be functional yet) https://github.com/YunoHost-Apps/polr_ynh

PR #2028, opened 2022-07-10
Add Authentik

CodeFreezr:
Authentik - Open source SSO-Solution with support for SAML2, OIDC, LDAP, SCIM & more


alexAubin:
This doesnt sounds super relevant to add to Yunohost @_@ 

I mean Yunohost already has one SSO, and it's not like you could install an SSO "as an app" and it would magically interfaces with other apps, there's usually too much tight-coupling to configure

ericgaspar:
I agree with @alexAubin

PR #2029, opened 2022-07-11
correct SFTP/SSH setup info

licynthiax:
Hi! I ran into some trouble getting SFTP set up with SSH properly, so here are some edits that would've helped me out. (The French document only corrects the webapp user's username; I barely know French)

Thanks!


alexAubin:
Cheers !

PR #2030, opened 2022-07-11
jitsi: link to electron app

csett86:
The Jitsi Desktop is a legacy SIP app not working with Jitsi Meet, replace with the Electron-based app.


csett86:
@ericgaspar I kindly ask you to review and merge this. If you need any further details, I am one of the maintainers of the Jitsi Electron App, see release history: https://github.com/jitsi/jitsi-meet-electron/releases

ericgaspar (commented):
Do we need to remove Jitsi Meet Desktop entry? Jitsi Meet Desktop and Jitsi Meet Electron are two different packages.

csett86:
Yes, the Jitsi Desktop entry (not Jitsi Meet Desktop) is an old Java-based SIP/XMPP client that does not work with Jitsi Meet. So linking it here (https://yunohost.org/de/app_jitsi) gives the wrong impression that it works with the yunohost Jitsi Meet app one just installed. Thats why it I strongly suggest to replace it with the electron app that actually works with Jitsi Meet.

csett86:
Thank you for merging @alexAubin

PR #2065, opened 2022-08-14
Add Flus(io) to wishlist

alexAubin:
This sounds like a very nicely designed app with special attention given to the UI/UX, a fresh look on RSS-reader and the like

On the other hand, the dev seems to be struggling to try to create a viable economic model based on selling Flus as SaaS : https://flus.fr/tarifs 

Which makes me think we should probably add a `support` key in the manifests to store and URL where people can support their favorite apps, and advertise those links from our webadmin

Edit: https://github.com/YunoHost/apps/commit/0ab409fbd0f2157cd23e5ee3ef9a780e00c0dabf as a followup


ericgaspar:
I created the repo : https://github.com/YunoHost-Apps/flusio_ynh. It is not working yet, I getting a white page. Something to do with the CLI command...

ericgaspar:
Closing with https://github.com/YunoHost/apps/pull/1496

PR #2092, opened 2022-09-15
Add headscale/headscale-ui

renne:
An open source, self-hosted implementation of the Tailscale Mesh-VPN control server.


tituspijean:
Wow it seems it has gotten much simpler to deploy. I'll definitely investigate it. :)

PR #2096, opened 2022-09-18
Italian translation for evaluate.md

noccio:
I have an italian translation for the file evaluate.md: where could I send it?


tituspijean:
Invalid PR, see my instructions in the chatroom.

PR #2158, opened 2022-12-02
Remove an indefinite article

Zoomerhimmer:
To make the English sound more natural.


Zoomerhimmer:
Optionally one might say: "It is a free software package." But 'software' alone doesn't sound like it can be counted in English: ~~1 software~~, ~~2 softwares~~, etc.

PR #2168, opened 2022-12-20
Explain SSH port alteration cannot be tested

Gredin67:
I ran into the issue that I could not connect with SSH with the new port while keeping connection with the old port, because new port was only reachable after a server reboot, hence my suggestion. But maybe reloading Fail2Ban would be sufficient?


ericgaspar (commented):
Isn't there a contradiction where it says how to test the new SSH configuration and say changing the SSH port cannot be tested?

tituspijean:
I have just tested the following:

1. Log in with SSH, keep the terminal open
2. Change the SSH port
3. Log in with SSH with the new port (it works, and the old terminal still works too).
4. Change the port back?
5. Both still work.

Closing. Feel free to ask for support on the usual channels. :)

PR #2261, opened 2023-05-07
added postifx restart

jakobkilian:
Hope that is fine and intended... Took me some time to realise


alexAubin:
Unfortunately this shouldnt be needed, it's already automatically restarted by YunoHost when changing the relay settings ... https://github.com/YunoHost/yunohost/blob/dev/hooks/conf_regen/19-postfix#L84 :neutral_face:

PR #2266, opened 2023-05-16
fix iptables example to forbid outgoing port 25

aya:
iptables rule target should be REJECT instead of ACCEPT.


alexAubin:
Hmmmkay sure, as far as I understand it's only meant to be an example of what hooks are and how they can be used, not a refined firewall, but why not i guess ...

PR #2271, opened 2023-05-23
Update BeatBump Official Domain

SilverSilt:
The domain beatbump.ml went down due to registration issues. The new domain is beatbump.io

See https://github.com/snuffyDev/Beatbump/pull/169/commits/df274b843aa938caea7a1371026d6cdee25920b9

https://github.com/snuffyDev/Beatbump/issues/168


ericgaspar:
Here we go. The upstream app is not well documented... So there may be some extra work to do to make it fully functional
https://github.com/YunoHost-Apps/beatbump_ynh

PR #2278, opened 2023-06-08
Modify occ command to avoid a PHP error

oleole39:
Cf. https://forum.yunohost.org/t/solved-using-occ-command-on-nextcloud/19788/5
Same error was met using YNH 11.1.20 & nextcloud-ynh 26.02~ynh2


oleole39:
PR closed as the commit is already in #2279 (git mess)

PR #2290, opened 2023-06-18
Update to the Try Yonohost admin credentials

johnkennyuk:
today I tried out the demo instance, the user ui creds worked but not the admin ui creds,  I managed to work out the correct creds and decided to edit the docs with this pull request.


alexAubin:
Uuuuuh, I'm unable to use those ... The credentials still are "demo" and "demo" on my side ...

PR #2292, opened 2023-06-18
Commit to fix the try yonohost admin creds in all lang md files

johnkennyuk:
Commit to fix the try yonohost admin creds in all lang md files


alexAubin:
... Still confused because these does not work ...

PR #2295, opened 2023-06-22
Update apps_wishlist.md

Philtouet:
Ajout de Projeqtor. Merci de jetre un œil sur cette super application de gestion de projet  libre et gratuite.
[projeqtor.org](https://www.projeqtor.org/fr/)


alexAubin:
Been trying for like 15 minutes to find where the hell the code is hosted and whichever install instruction, the only thing I found is a Dockerfile on a random github project which may be reverse-engineerable to find out some sourceforge repository ... Doesn't look really self-hosting friendly

PR #2299, opened 2023-07-15
Add Immich to apps_wishlist.md

kingofthejerichostone:


ericgaspar:
Immich is already in the catalog but is not working... https://github.com/YunoHost-Apps/immich_ynh

PR #2316, opened 2023-08-08
Add permission check for hooks directory

CodeShakingSheep:
When I used hooks for the first time in yunohost I stumbled upon missing directory permissions. Just added this information in to the docs for hooks. Also corrected some minor incorrect spellings on the way.


zamentur:
777 permissions are a bit too much. When an app is setup, the hooks dir are created with 755 and root:root. And file inside with 644.

PR #2317, opened 2023-08-09
Remove fittrackee from wishlist

CodeShakingSheep:
[Fittrackee](https://github.com/YunoHost-Apps/fittrackee_ynh) is already packaged and currently has integration level 7. I just spotted it on the wishlist and removed it.


Tagadda:
o/

PR #2320, opened 2023-08-19
Add OpenWRT to wishlist

GeraldSoellinger:
Would be cool to run OpenWrt as VM on a Yunohost managed machine.


alexAubin:
Could you elaborate how "cool" it would be ..?

As far as I know, OpenWRT is a router distribution ... it feels pretty inappropriate to run it inside a VM, or you'll have to do a whole bunch of iptables trick on the host, and even then probably will experience a bunch of mystic issues.

GeraldSoellinger:
I use "[PC engines](https://pcengines.ch/)" hardware with OpenWRT installed as my network equipment. 
Since it has plenty of resources left I would like to install other services too, but some services are not available or easy to maintain on OpenWRT distribution. Many services I like are handy to maintain with yunohost. 
As I don't know how to configure debian as a wifi access point for example I need openWRT for that.

alexAubin:
If your use case is "wifi access point", we have an hotspot app for that

Still confused about how OpenWRT-in-a-VM is gonna be useful for anything practical ...

GeraldSoellinger:
okay, I gonna try the [wifi hotspot app ](https://github.com/YunoHost-Apps/hotspot_ynh) .
If it works out, you can forget about this request.

PR #2321, opened 2023-08-20
[enh] Add config panel docs

zamentur:
A docs for config panel.


zamentur:
SO i discovered a doc was writen by axo, i merged cause a user need a doc, but i believe axo's doc seems more better (maybe missing some element about getter / setter maybe)

PR #2331, opened 2023-09-11
fix missleading example for ynh_setup_source

mrflos:
there shouldn't be any trailing / for folders for ynh_setup_source


alexAubin:
Eeeeh it's an autogenerated doc ... you probably want to modify https://github.com/YunoHost/yunohost/blob/dev/helpers/utils#L78 instead

mrflos:
Why is autogenerated code available in the source ? I got a little confused, sorry..

Ok so https://github.com/YunoHost/yunohost/pull/1714 : from nobody, to yunohost-Apps member to core contributor in one day, this is the way!

tituspijean:
Superseded by https://github.com/YunoHost/yunohost/pull/1714

PR #2335, opened 2023-09-17
Belenios voting added to wishlist

stefanxyz:
We need secure and anonymous voting software where voters can trust that nobody was manipulating. My research came to conclusion, that Belenios with origins in Université de Lorraine, is most promising candidate. 
The survey platforms that are provided in YunoHost so far - ("LimeSurvey" and "Framaforms") - are rather "survey" but not "election" software. 
E.g. LimeSurvey admin can collect all data of the voter (IP, voting selection, e-Mail-address ...) and even vote in place of the original voter. Belenios focuses on maximum safety:
1) anonymous voting
2) all voters get their individual partition token, repetitive sending if someone did not get his token 3) verifiable: 
  a) voters can verify that their ballot was counted;
  b) voters can verify that there was no ballot stuffing
4) board of admins provides control that there is less influencing from the admin


stefanxyz:
Please consider to add Belenios Voting to the YunoHost wishlist.

ericgaspar (approved):

PR #2340, opened 2023-09-17
Update apps_wishlist.md

isAAAc:
add Yakforms to wishlist


alexAubin:
Eeeh I'm confused about this ... Framaforms is already packaged, and the plan is definitely to move to Yakforms, so it's not really about "adding the app to the wishlist", it's more about finalizing the existing PRs and ideally going through all the "app renaming" process ...

PR #2364, opened 2023-11-05
Update external_storage.md

pidlas:
C'est une commande bash plus courte que celle proposée initialement.


pidlas:
This is a minor modification in the documentation about 'Adding an external storage, 6. Automatically mount on boot'. I propose to get the UUID of the newly installed external HDD to use "lsblk -f" instead of "blkid | grep "/dev/YOUR_DISK1:"".

PR #2370, opened 2023-11-30
Update borgbackup.fr.md

supermamie:
Add an explanation to generate an archive and excluding a part of it


alexAubin:
Closing because it's been moved to https://github.com/YunoHost-Apps/borg_ynh/blob/master/doc/ADMIN.md and the `exclude` option is described in the section "Restoring the "source+config" of the app, and its data separately"

Sorry for the huge delay

PR #2392, opened 2024-02-04
Update xmpp.md

anubister:
Adding limitations and some workarounds


alexAubin (commented):
LGTM but ideally should be translated to french as well such that the two pages don't diverge too much

oleole39:
See #2412 for FR version

PR #2409, opened 2024-02-19
Warns that there’s no need to hide servers IP for lambda users

OniriCorpe:
Yesterday I saw a company selling servers with yunohost pre-installed (nice) but also boasting about offering VPNs "for privacy" because they hide the server's IP address
And I still find this fear inducing marketing disgusting


tituspijean:
Kinda related to https://github.com/YunoHost/issues/issues/2332 (for future follow-up)

> I saw a company selling servers with yunohost pre-installed

ah nice! Maybe we could consider adding them in the providers list? ...

> but also boasting about offering VPNs "for privacy" because they hide the server's IP address
> And I still find this fear inducing marketing disgusting

... maybe not xD

PR #2433, opened 2024-03-07
Fix legacy admin user mentions

tituspijean:
## PR checklist

- [x] I'm not doing a PR for an application, I promise, I know that this kind of changes must go directly into the app packages themselves


tituspijean:
Merging as is, german and italian translations need to be done. In the meantime I added warnings that the sections  are not up to date.

PR #2492, opened 2024-09-16
expanded note on bash variables during install script

oleole39:
Added details to the explanation of bash variables


pages/06.contribute/10.packaging_apps/20.scripts/scripts.md
```diff
@@ -64,7 +64,7 @@ Note that the scripts are run with the `set -eu` options (except for the remove
...
 
 - `$app` is the app ID. It will typically be the ID from the app's manifest.toml, for example `helloworld`, but will be `helloworld__2`, `__3` etc for multi-instance installs.
-- During install, answers to install questions are automatically available as bash variables. For example, the `$domain` setting corresponds to the `domain` question, same for `$prefered_pet`, etc... Note that - apart from special questions such as `init_main_permission` or user-provided passwords - they are also automatically saved as settings (cf next section).
+- During install, answers to install questions are automatically available as bash variables. Should thus be available as bash variables at least `$domain`, `$path` and `$init_main_permission` which contains the answers to the default install questions defined in the `manifest.toml`. Answer values associated to any custom install questions will be available the same way, e.g. `$prefered_pet` for the custom question `[install.prefered_pet]` that would be defined in the manifest. Note that - apart from special questions such as `init_main_permission` or user-provided passwords - they are also automatically saved as settings (cf. next section).
```
alexAubin, line 67:
```suggestion
- During install, answers to install questions are automatically available as bash variables. Typically, `$domain` and `$path` will contain the answers to the default install questions assuming they are defined in the `manifest.toml`. Answer values associated to any custom install questions will be available the same way, e.g. `$prefered_pet` for the custom question `[install.prefered_pet]` that would be defined in the manifest. Note that - apart from special questions such as `init_main_permission` or user-provided passwords - they are also automatically saved as settings (cf. next section).
```

(not every app has `domain`/`path` question, and `$init_main_permission` shouldnt have to be used by packagers, it's only meant to initialize the main permission which is handled by the core)

PR #2493, opened 2024-09-16
detailed 'accept' parameter of file type question  

oleole39:
improved description and added examples.
The first example was successfully tested, as opposed to `[".png"] which was not working.
The 2 last examples have been copied from Mozilla's doc but have not been tested with YNH... would it rather be ["image/png", "image/jpeg"] ?

### DIFF
--- a/pages/06.contribute/15.dev/03.forms/forms.md
+++ b/pages/06.contribute/15.dev/03.forms/forms.md
@@ -130,7 +130,7 @@ File question
 
 | Property | Description  | Example |
 |----------|--------------|---------|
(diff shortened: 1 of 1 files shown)


pages/06.contribute/15.dev/03.forms/forms.md
```diff
@@ -130,7 +130,7 @@ File question
 
 | Property | Description  | Example |
 |----------|--------------|---------|
-| `accept`    | Same format than HTML file input |         |
+| `accept`    | Accepted MIME types, as in [HTML implementation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept) | `["image/png"]` or `["image/*"]` or `["image/png, image/jpeg"]` |
```
alexAubin, line 133:
```suggestion
| `accept`    | Accepted MIME types, as in [HTML implementation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept) | `["image/png"]` or `["image/*"]` or `["image/png", "image/jpeg"]` |
```

PR #2502, opened 2024-10-07
Add instructions to check locally whether autoupdate strategy is working

oleole39:
Add instructions to check locally whether autoupdate strategy is working


alexAubin:
(as stated on top) This file is autogenerated from the python code : cf https://github.com/YunoHost/yunohost/blob/dev/src/utils/resources.py#L421

oleole39:
> (as stated on top) This file is autogenerated from the python code : cf https://github.com/YunoHost/yunohost/blob/dev/src/utils/resources.py#L421

Right, not the first time I get mistaken here...
Note that the `Edit` link at the top of the page is kind of misleading as it points to https://github.com/yunohost/doc/blob/master/pages/06.contribute/10.packaging_apps/10.manifest/10.appresources/packaging_app_manifest_resources.md
 
editdoc

alexAubin:
Sure, if you find a way to disable edit button on specific page then we can disable it on this on and the helper doc too ...

PR #2507, opened 2024-10-13
Create installing_debian.fr.md

scls19fr:
## Problem

- Only english documentation was available

## Solution

- Translation to french

## PR checklist

- [X] PR finished and ready to be reviewed


zamentur (changes requested):
Thanks for your contribution.

I think this instructions should be included in the install tutorial directly. Maybe by giving more explanation and making a link here https://github.com/YunoHost/doc/blob/3a2c392bcd67e754fb163014256544d8c1f63e11/pages/02.administer/10.install/install.fr.md?plain=1#L471

The link could be on the same install tutorial but with other params ( `https://yunohost.org/fr/install/hardware:regular_debian` for example )

PR #2557, opened 2025-02-09
Detail a bit more how an app should SSO-logout

fflorent:
## Problem

- I had difficulties to know how to make my app logout properly as detailed here: https://github.com/YunoHost-Apps/grist_ynh/issues/64#issuecomment-2645938287

## Solution

- I propose to mention that the app has to redirect to the main domain to make this route work: `/yunohost/sso/?action=logout`

## PR checklist

- [x] PR finished and ready to be reviewed


fflorent:
Converting to draft, as it does not work like described anymore since Yunohost 12.

alexAubin:
Discussed on the chat the other day, no time to iterate on this right now but the correct URI is `/yunohost/portalapi/logout?referer_redirect` (which will redirect to the referer of the redirect after being logged out)

The domain part should be the "topest domain" for the app's domain ... or maybe it does work with the app's (sub)domain too ? Needs to be tested ...

zamentur:
I tried this redirect api endpoint with nextcloud, but sadly it doesn't work due to no-referer policy /o\

https://github.com/YunoHost-Apps/nextcloud_ynh/blob/2414a029fddf01a5e5b2a53878d02f24441a90c3/conf/nginx.conf#L67

fflorent:
> The domain part should be the "topest domain" for the app's domain ... or maybe it does work with the app's (sub)domain too ? Needs to be tested ...

It did not work with a subdomain. I don't know how to get the topest domain properly.

fflorent:
I achieved something using this command: `yunohost --json domain info $domain | jq -r '.topest_parent // "'$domain'"'`. Does this look accurate?

PR #2572, opened 2025-03-15
Reuse existing translations

Thovi98:
Most of the existing translations is now ported on Docusaurus.

When the pages were to different (a to old copy of the page), I’ve not taken the translation (it’s for 2-3 pages max).

## PR checklist

- [x] PR finished and ready to be reviewed
- [x] I agree that my pr will be squashed (I’ve learned a bit since last pr ^^)


Salamandar:
You duplicated english pages and that's not needed : Docusaurus fallbacks gracefully to the default locale when pages don't exist.
We should only create translated pages when they are really translated.

Thovi98:
Hm interesting, I’m following the [Docusaurus guide](https://docusaurus.io/fr/docs/i18n/git#initialize-the-i18n-folder).
If you feel we should still remove the non translated ones (what would be the benefit?), I’ll remove after incorporating all existing translations.

Salamandar:
Thanks !!!

Like the previous PR though, github doesn't understand the merge conflicts so I squashed / rebased your branch.

I left you as git author of course !

I also just removed the commit that deleted the grav pages : I saw some missing stuff (at least in the russian translation)

PR #2575, opened 2025-03-22
Some Docusaurus fixes

Thovi98:
## Problem

- *Description of why you made this PR, what is its purpose*
image

## Solution

Change some image urls (I’ve changed them in all _pages_ also, so it’s already done)

I’ve also added some footer links which are more useful I think ;)

(Hopefully my learning to git is in a good wave and there must be less issues merging my branch haha)

## PR checklist


Thovi98:
By the way @Salamandar 
image

I don’t understand why some files are duplicated like that?

Salamandar:
> By the way @Salamandar 
> I don’t understand why some files are duplicated like that?

Yeah that is my mistake, i don't really understand what I did and I think I will fix this later tonight or tomorrow

PR #2599, opened 2025-05-07
A bit more detailed custom config getter for YAML format

olberger:
The way to generate YAML content is a bit obscure whenever one is confronted to it in the first place. Hope these tiny bits make it more easy to understand

## PR checklist

- [X] PR finished and ready to be reviewed


Salamandar:
Cherry-picked. Thanks !

PR #2603, opened 2025-05-22
add pr preview

Thovi98:
## Problem

- *Description of why you made this PR, what is its purpose*

## Solution

## PR checklist

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,68 @@
+name: Preview Docs
+
+on:
+  pull_request_target:
+    branches: [master]
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+  pull-requests: write
+
+concurrency:
+  group: preview-${{ github.head_ref }}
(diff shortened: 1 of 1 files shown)


Thovi98:
I don’t know if this is useful for you (I enjoyed learning how Github Actions works).
On every pr, the docs will be deployed on `yunohost.github.io/docs`. 
Unfortunately, I didn’t succeed to make it so that every pr is located in its own subpath (e.g. `yunohost.github.io/docs/pr-2603`). IIUC for this to work we have to use some third-party tools like vercel or cloudflare pages.
So currently, the link will be updated with the latest changes on a pr.

For the pr to work, you have to create an environment called `preview` with write permissions.

Salamandar:
I've had to use `actions/deploy-pages` at $dayjob and came to the conclusion that it's basically frozen crap. You can't distribute on a subpath (it overwrites *everything* on the pages instead of just the subpath). we might need to git clone / checkout gh-pages / commit / push manually instead.

PR #2617, opened 2025-06-26
Document supported CPU archs

oleole39:


docs/packaging/10.manifest/10.resources.mdx
```diff
@@ -379,7 +379,8 @@ Various options are available to accomodate the behavior according to the asset
     autoupdate.strategy = "latest_github_tag"
 ```
 
-Or more complex examples with several element, including one with asset that depends on the arch
+Or more complex examples with several element, including one with asset that depends on the CPU architecture.
```
alexAubin, line 382: This page is autogenerated from resources.py i.e. https://github.com/YunoHost/yunohost/blob/dev/src/utils/resources.py#L337 x_x

oleole39: Also this section would need to be updated with appropriate instructions: https://doc.yunohost.org/dev/doc/
I wouldn't mind doing it, but I would need first to understand the objective of this new doc setup which from documentation's contributor's point of view is a bit more complicated.

Otherwise, feel free to close this PR.

alexAubin: >However this is kind of new I believe - could related instructions be added to the doc repo's README file?

Idk it's been like this since the existence of resources i think ... i feel like it's the 5th or 6th time somebody tries to make a PR on such a file, I don't know how to make it more obvious, it's explained at the very top of the file ...

alexAubin: >from documentation's contributor's point of view is a bit more complicated.

How so ? Personally I'm not a huge fan of this "how to write doc" section ... The doc about how to build docusaurus locally is more targeted at people maintaining the doc and should be in the doc's README imho ... And ideally we should have Dockerfile for this but that's a yakafokon.

The `https://doc.yunohost.org/dev/doc/apps_documentation` page feels unecessary and was there because of legacy story about some old app doc page 

The markdown guide feels out of place, there's already a gazillion markdown tutorials on the internets

And I don't know what to think about the "Contribute via Github" ...

oleole39: > I don't know how to make it more obvious, it's explained at the very top of the file ...

Granted, I had not checked the top of the file... I guess I am still too used with the old way of editing the doc.
However to be fair, the top of the file links to `doc/scripts/resources_doc_generate.py` and from there it requires reading and understanding the code to eventually get to I have should edited  `/src/utils/resources.py`. 
Not obvious enough to my mind, I think there should be:
1. An index listing the special pages in the dedicated section (https://doc.yunohost.org/dev/doc/#special-pages) showing the link to be modified or at least which kind of special pages can be found in which folder.
2. Appropriate links for "Edit this page" link on each doc page (e.g. https://doc.yunohost.org/packaging/manifest/resources/#:~:text=Edit).
3. As you said, most of the [Write documentation page](https://doc.yunohost.org/dev/doc/) is actually more about "Deploying documentation" and will be meaningful for a limited number of people, whereas I think "Writing/Editing documentation" should be broadly open to contributors as you don't need to be a seasoned dev to do that and you may often have ideas while experimenting a trouble where someone more familiar to how YunoHost works would not have thought of explaining. So IMO we should on the one hand a "Deploy documentation" page containing the current "Docusaurus" section, and a "Write documentation" containing:
    - brief explanation of the technology (Docusaurus + custom scripts).
    - explanation on how to contribute (via Github if suggestion 2 above is actually followed, or/and the index suggested in 1 above)
4. a brief explanation of the technology and links to "Write documentation" and "Deploy documentation" pages in [the repo README](https://github.com/YunoHost/doc/blob/main/README.md). 
  
   

> The https://doc.yunohost.org/dev/doc/apps_documentation page feels unecessary and was there because of legacy story about some old app doc page

Agree. I don't know about the legacy part, but yes I would find it better located as a section in https://doc.yunohost.org/dev/doc rather than in its own page as it is now.

> The markdown guide feels out of place, there's already a gazillion markdown tutorials on the internets

True, but somehow it doesn't hurt and it is still relevant so I would keep it.

> ideally we should have Dockerfile for this but that's a yakafokon.

I'm sure you meant "a YNH package" ;)

oleole39:
Discussion can continue on #2667

PR #2619, opened 2025-07-06
Update raspberry_pi.mdx

Marc-Flender:
Add informations for RaspberryPi 2.
For example for those who crash their install...

## Problem

- *I had to reinstall on raspberryPi2, but the img files did not comply with my machine*

## Solution

## PR checklist

### DIFF
--- a/docs/admin/05.get_started/20.install_on/raspberry_pi.mdx
+++ b/docs/admin/05.get_started/20.install_on/raspberry_pi.mdx
@@ -16,6 +16,12 @@ import YunoHostImagesList from '@site/src/components/YunoHostImagesList';
 
 ## <FAIcon icon="fa-list-alt"/> Pre-requisites
 
+:::note
+If you wish to install it on a Raspberry Pi 2, you can either
+  - follow this guide but with older image from Debian Buster ([Download here](https://repo.yunohost.org/images/attic/yunohost-buster-4.1.7.1-rpi-stable.img.zip)) and then  [upgrade the system](/admin/tutorials/upgrade)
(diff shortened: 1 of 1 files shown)


docs/admin/05.get_started/20.install_on/raspberry_pi.mdx
```diff
@@ -16,6 +16,12 @@ import YunoHostImagesList from '@site/src/components/YunoHostImagesList';
 
 ## <FAIcon icon="fa-list-alt"/> Pre-requisites
 
+:::note
+If you wish to install it on a Raspberry Pi 2, you can either
+  - follow this guide but with older image from Debian Buster ([Download here](https://repo.yunohost.org/images/attic/yunohost-buster-4.1.7.1-rpi-stable.img.zip)) and then  [upgrade the system](/admin/tutorials/upgrade)
```
alexAubin, line 21: I don't think we want to advertise a buster image to people ... with Trixie around the corner it's going to be **three** major versions behind ...

Marc-Flender: Hi Alex,
I uderstand your answer.
The problem is that when SD card crashes, you don't change the machine, just the SD Card, and you have to keep your Pi2 instead of buying a new one.
That happened to me this week, and i tried to reinstall. 
So i used Yunohost Documentation, and i first did not understand the Raspberry page was dedicated to Pi >3. So i flashed my Pi2 with the wrong image
Then i searched for a correct image in Yunohost Attic repository, but there is no mention of compatibility with Pi models.
So i tried several ones, up to version 4. 

So i think i have wasted about 2 or 3 hours to find the correct image.

Then, I think that a mention like the one i suggest, may be useful for users, no to waste time as i did, even more if they are nos specialists as you may be.
Don't you think ?

Issue #2620, opened 2025-07-12
Broken link in yunohost admin to doc for registrar


fflorent:
## The problem

When you go to:
1. The admin panel (`/yunohost/admin`);
2. The domains list, and click on the principal one (or whatever)
3. The DNS tab;

You have the following text (`domain_dns_registrar_supported`):
>    YunoHost automatically detected that this domain is handled by the registrar **{registrar}**. If you want, YunoHost will automatically configure this DNS zone, if you provide it with the appropriate API credentials. You can find documentation on how to obtain your API credentials on this page: https://yunohost.org/registar_api_{registrar}. (You can also manually configure your DNS records following the documentation at https://yunohost.org/dns )

[the corresponding source](https://github.com/YunoHost/yunohost/blob/e3c83f90af1406007e73b87ab4ecffc1fdeb3fad/locales/en.json#L431)

However, the `/registrar_api_{registrar}` seems to now be a dead link.

## Proposed solution

My guess is that we may restore this route to redirect to the page of the new documentation.

Thovi98:
I think I’ve fixed it in https://github.com/YunoHost/doc/pull/2606

PR #2640, opened 2025-08-22
documentation overhaul for DNS and domains topics

MrLixm:
Hello, this PR introduce an overhaul of the documentation about domains and DNS.

I initially went for minimal changes but got a bit obsessed and end-up going full in @_@

DNS and domain were probably the most confusing topics I had when using YunoHost so I thought it would be worth having an extra-polished documentation for this topic. This means I am not an expert in those domains and relied mainly on external sources to verify my understanding.

The overall change direction is trying to regroup information and avoid having it scattered on different pages.

I am open to discuss the changes using other communication medium as I'm aware the changes might be hard to review.

Let me know what you think and what could be done better,
Cheers !
Liam.

## what changed

✏ **(modified) `get_started/install_on`**

improved the “Main domain” section, added a Tabs widget to present the 3 options.

✏ **(modified) `/get_started/post_install/dns_config`**

the content of this page was fully moved to `/domains/dns_configuration`

instead we simply redirect the user to the necessary steps he might have to perform

✏ **(modified) `/domains/index`**

- overall polish and simplification
- new screenshot added (more recent YunoHost version)
- added missing 4th subdomain option in the Tabs widget
- added “Installing an application on a domain” section
- added “Changing an application assigned domain.” section

🗑 **(removed) `/tutorials/domains/dns_subdomains`**

this page was removed and merged with the new `/domains/dns_configuration` and `/domains/index`

🗑 **(removed) `/tutorials/domains/dns_nohost_me`**

this page was moved to `/domains/dns_nohost_me`

✨ **(added) `/domains/dns_nohost_me`**

this page was originally `/tutorials/domains/dns_nohost_me`

- overall simplication
- completely changed “Changing your free domain” and “Retrieving an existing free domain” sections

✨ **(added)  `/domains/dns_configuration`**

this page was originally `/tutorials/domains/dns_subdomains`

- added "A brief explication of domain names” section
- added “A brief explication of DNS records.” section
- used the `autohotkey` syntax highlighting for code blocks (works best with zone file syntax)
- “Recommended DNS configuration”:
    - removed the table as this created redundant information and we instead explain the code block structure in the new section.
- “Subdomains”:
    - this was partially retrieved from `/tutorials/domains/dns_subdomains`
    - added another example if your apex domain is already in use and you just want to set new subdomains
- should fix #2581 

## need input

1. I added redirects for page which have been removed, is it fine ?
2. I didn’t update the i18n pages because I don’t know how they work, is it possible to have an explanation on how to update those ?
3. `/admin/domains/dns_nohost_me#retrieving-an-existing-free-domain`
when a user forgot is free domain password, is it okay to mention they must ask a reset on the forum thread ?
4. Should we refer YunoHost's domains as “free domains” / “automatic domains” or even “dynDNS domains” ?
5. `/admin/domains/index#changing-an-application-assigned-domain`
Is the workflow to reinstall an app correct ? I’m not sure if we shoudlo really reinstall the app before restoring the backup. (this are the pre-pr instructions).
6. The recommended DNS records include wildcard records. Should we add a note explaining wildard record ? Are they really necessary ?
7. In the YunoHost web interface > Domains > {yourdomain.tld} > DNS, there is 2 documentation links that are broken. Should we add redirections for those too ? (related to #2620 )
(upload screenshot)
8. with a fresh git clone, I had error building the doc. I had to fix `docs\admin\05.get_started\10.providers\05.registrar\_category*_*.yaml` by adding a `Label: Registrar` key inside. this change have not been commited but maybe should be ? I don’t know how  category.yaml works.

## references

- https://en.wikipedia.org/wiki/Domain_name
- https://en.wikipedia.org/wiki/Subdomain
- https://en.wikipedia.org/wiki/Zone_file
- https://stackoverflow.com/questions/55461299/what-is-the-correct-term-for-apex-naked-bare-root-domain-names
- https://serverfault.com/questions/955112/what-does-mean-in-a-hostname-in-dns-configuration

### DIFF
--- a/docs/admin/25.domains/index.mdx
+++ b/docs/admin/25.domains/index.mdx
@@ -115,13 +115,9 @@ You can change the application domain:
 - through the web interface > `Applications` > `{app name}` > `Operations`
 - through the command line with `yunohost app change-url`
 
-Some applications don't allow domain change. The best solution is to reinstall the application:
-
-- backup the application data (`yunohost backup create --apps ...`)
-- uninstall the application
-- reinstall the application to the desired domain
-- restore the backup 
-
+:::note
+Some applications do not allow you to change domains. In general, it is still possible to make the change manually, but this operation requires a variety of system administration skills.
+:::
 
 ## DNS configuration
 


zamentur:
> 1. I added redirects for page which have been removed, is it fine ?
Yes thanks

> 2. I didn’t update the i18n pages because I don’t know how they work, is it possible to have an explanation on how to update those ?
For example, french translations for `/docs/admin/25.domains` are in `/i18n/fr/docusaurus-plugin-content-docs/current/admin/25.domains`. All the .mdx should be translated...

We have no easy way to do it, we think put in place some .po mechanism, but it's not the case for now.

Did i answer your question ?

> 3. `/admin/domains/dns_nohost_me#retrieving-an-existing-free-domain`
> when a user forgot is free domain password, is it okay to mention they must ask a reset on the forum thread ?
Yes for domain set before yunohost 11.2 we accept request for reseting a domain.

> 4. Should we refer YunoHost's domains as “free domains” / “automatic domains” or even “dynDNS domains” ?
Where ?

> 5. `/admin/domains/index#changing-an-application-assigned-domain`
> Is the workflow to reinstall an app correct ? I’m not sure if we shoudlo really reinstall the app before restoring the backup. (this are the pre-pr instructions).
I made a commit for this

> 6. The recommended DNS records include wildcard records. Should we add a note explaining wildard record ? Are they really necessary ?
This is a good question, indeed wildcard record has advantage and pitfals that are not explained here. Specifically if users mistype a subdomain...

> 7. In the YunoHost web interface > Domains > {yourdomain.tld} > DNS, there is 2 documentation links that are broken. Should we add redirections for those too ? (related to #2620 )
> (upload screenshot)
Yes, feel free to do it, thanks !

> 8. with a fresh git clone, I had error building the doc. I had to fix `docs\admin\05.get_started\10.providers\05.registrar\_category*_*.yaml` by adding a `Label: Registrar` key inside. this change have not been commited but maybe should be ? I don’t know how  category.yaml works.
I don't know myself sorry.

PR #2748, opened 2026-02-18
Fix `packages` property's example syntax

oleole39:
So far 'packages' property in `manifest.toml` is meant to be formatted as a comma-separated string. Cf. https://github.com/search?q=org%3AYunoHost-Apps+path%3Amanifest.toml+resources.apt&type=code


oleole39:
Err... actually we also find the format I removed in this PR: https://github.com/YunoHost-Apps/pihole_ynh/blob/22e6708e48f7f61dd67a7fbf549894c616369b4b/manifest.toml#L89

Both are working I guess. Do we want to migrate to a specific format or should we tell about the two formats in the doc ?

PR #2752, opened 2026-03-25
Update config panel and question types

spicy-catch-268:
## Problem

Doc not up to date, missing 'display_text', 'range' and 'tags-select' question types.

## Solution

Complete doc. **Need help to complete respective sections.**

## PR checklist

### DIFF
--- a/docs/dev/50.packaging/60.advanced/20.config_panels.mdx
+++ b/docs/dev/50.packaging/60.advanced/20.config_panels.mdx
@@ -157,7 +157,7 @@ ynh_app_config_run $1
 
 #### Custom getters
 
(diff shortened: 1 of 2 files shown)


spicy-catch-268:
Also, in [doc](https://doc.yunohost.org/fr/dev/core/forms#string-string), string fields aren't supposed to support "choices" property even if they do. When set, the fields are rendered as select fields instead.

Projects using black magic:
- healthchecks_ynh
- django_example_ynh
- django-fmd_ynh
- django-for-runners_ynh
- django-fritzconnection_ynh
- pyinventory_ynh

Maybe this is for compatibility? If so, shouldn't maintainers of those projects be notified to update? If not, shouldn't the doc reflect this behavior ?

PR #2777, opened 2026-08-16
update the raspberry pi local hostname used on first boot

mexon:
## Problem

- I was trying to set up yunohost on a raspberry pi, and tore my hair out until I realised that `yunohost.local` was a completely different machine.  When the current raspberry pi image (v12.0.16) boots, it has a hostname of `rpi`, not `yunohost`.

## Solution

- Update the documentation to point to the correct local hostname.

## PR checklist

### DIFF
--- a/docs/admin/15.command_line.mdx
+++ b/docs/admin/15.command_line.mdx
@@ -31,8 +31,8 @@ If you connected with the administration account and would like to become `root`
 
 If you are **installing at home** (e.g. on a Raspberry Pi or OLinuXino or old computer):
 
-- you should be able to connect to your server using `yunohost.local` (or `yunohost-2.local`, depending on how many servers are on your network).
-- if `yunohost.local` and the like do not work, your need to [find out the local IP of the server](/admin/get_started/post_install/find_ip).
+- you should be able to connect to your server using `rpi.local` (or `yunohost-2.local`, depending on how many servers are on your network).
+- if `rpi.local` and the like do not work, your need to [find out the local IP of the server](/admin/get_started/post_install/find_ip).
 - if you installed a server at home but are attempting to connect from outside your local network, make sure port 22 is correctly forwarded to your server.
 
 If your server is a remote server (VPS), your provider should have provided you with the IP address of the machine
@@ -54,7 +54,7 @@ ssh username@11.22.33.44
 ssh username@your.domain.tld
 
 # using the local domain name instead of the IP (for local access)
-ssh username@yunohost.local
+ssh username@rpi.local
 
 # if you changed the SSH port
 ssh -p 2244 username@your.domain.tld


tituspijean:
Thank you! That actually looks like a bug. Let us investigate. 

If you can, can you share the logs of the `yunomdns` service? You can anonymize any new domains, I'm only interested in its attempts to broadcast `yunohost.local` (or `rpi.local` here)

mexon:
@tituspijean OK, I see the problem.  This is from a fresh install:

```
root@rpi:~# journalctl --unit yunomdns
Mar 06 14:56:23 rpi systemd[1]: Started yunomdns.service - YunoHost mDNS service.
Mar 06 14:56:29 rpi yunomdns[761]: Uh oh, yunohost.local already exists on the network...
Mar 06 14:56:29 rpi yunomdns[761]: Adding yunohost-2.local with addresses ['192.168.2.119', 'fd6e:637c:3bbb:1:ba27:ebff:febe:1d4d'] on interface eth0
Mar 06 14:56:29 rpi yunomdns[761]: Registering...
Mar 06 14:56:29 rpi yunomdns[761]: Registered. Press Ctrl+C or stop service to stop.
```

There is in fact another yunohost raspberry pi on the network, since years (but also bookworm).  I usually ssh to it as `hs1.local` - "home server 1", I was always planning to add more.  I did not expect that box to continue occupying `yunohost.local` once I'd configured it with a different name.  I also do not expect the new box to be claiming both `yunohost.local` and `rpi.local` at the same time.

I'm actually doing this because I wanted to experiment with two raspberry pis using wireguard to talk to each other.  So it's a bit of a pain that there's poor support for multiple such machines on the same network.

But clearly the intent of this PR is wrong, so i guess I'll close it.

tituspijean:
> I did not expect that box to continue occupying yunohost.local once I'd configured it with a different name.  

It is, however, the documented behavior. (cf. https://doc.yunohost.org/en/admin/domains, "local domains" tab.

>  I also do not expect the new box to be claiming both yunohost.local and rpi.local at the same time. 

It is unclear to me where this `rpi.local` is coming from. Did you set it up?

> So it's a bit of a pain that there's poor support for multiple such machines on the same network.

I think the yunomdns service gracefully switching to `yunohost-2.local` is enough support to at least let you reach your server once before adding new local domains. What more support to you require?

mexon:
Closing PR, since `yunohost.local` really should be the definitive address to use.

mexon:
@tituspijean yeah, you're right, I not only didn't pick up all of the places in the documentation it suggested `yunohost-2`, I also didn't proofread my own PR properly.  I was stuck on the "ssh" section, since that's what I really wanted to do.  Sorry for the noise.

tituspijean:
No worries. :) You are most welcome on the support channels or the forum if needed. :)

